PR #17, opened 2020-11-23
Synchronize code base with upstream PcapPlusPlus

weyrick:
Fixes issue #17 

- fork DnsLayer to NS1 specific version
- fix upstream API changes, mostly with IP addresses and switch from timeval to timespec

### DIFF
--- a/src/geoip.cpp
+++ b/src/geoip.cpp
@@ -18,64 +18,11 @@ GeoDB::~GeoDB()
     MMDB_close(&mmdb);
 }
 
-std::string GeoDB::getGeoLocString(const uint32_t ip_address) const {
+std::string GeoDB::getGeoLocString(const struct sockaddr *sa) const {
 
     int mmdb_error;
-    struct sockaddr_in sa;
-    memset(&sa, 0, sizeof(sa));
-    memcpy(&sa.sin_addr, &ip_address, sizeof(sa.sin_addr));
-    sa.sin_family = AF_INET;
-    MMDB_lookup_result_s lookup = MMDB_lookup_sockaddr(&mmdb, (sockaddr *const)&sa, &mmdb_error);
-    if (mmdb_error != MMDB_SUCCESS || !lookup.found_entry) {
-        return "Unknown";
-    }
 
-    return _getGeoLocString(&lookup);
-
-}
-
-std::string GeoDB::getGeoLocString(const uint8_t* ip_address) const {
-
-    int mmdb_error;
-    struct sockaddr_in6 sa;
-    memset(&sa, 0, sizeof(sa));
-    sa.sin6_addr.s6_addr[0] = ip_address[0];
-    sa.sin6_addr.s6_addr[1] = ip_address[1];
-    sa.sin6_addr.s6_addr[2] = ip_address[2];
-    sa.sin6_addr.s6_addr[3] = ip_address[3];
-    sa.sin6_family = AF_INET6;
-    MMDB_lookup_result_s lookup = MMDB_lookup_sockaddr(&mmdb, (sockaddr *const)&sa, &mmdb_error);
-    if (mmdb_error != MMDB_SUCCESS || !lookup.found_entry) {
-        return "Unknown";
-    }
-
-    return _getGeoLocString(&lookup);
-}
-
-std::string GeoDB::getGeoLocString(const in_addr* in_addr) const {
-
-    int mmdb_error;
-    struct sockaddr_in sa;
-    memset(&sa, 0, sizeof(sa));
-    sa.sin_addr = *in_addr;
-    sa.sin_family = AF_INET;
-    MMDB_lookup_result_s lookup = MMDB_lookup_sockaddr(&mmdb, (sockaddr *const)&sa, &mmdb_error);
-    if (mmdb_error != MMDB_SUCCESS || !lookup.found_entry) {
-        return "Unknown";
-    }
-
-    return _getGeoLocString(&lookup);
-
-}
-
-std::string GeoDB::getGeoLocString(const in6_addr* in_addr) const {
-
-    int mmdb_error;
-    struct sockaddr_in6 sa;
-    memset(&sa, 0, sizeof(sa));
-    sa.sin6_addr = *in_addr;
-    sa.sin6_family = AF_INET6;
-    MMDB_lookup_result_s lookup = MMDB_lookup_sockaddr(&mmdb, (sockaddr *const)&sa, &mmdb_error);
+    MMDB_lookup_result_s lookup = MMDB_lookup_sockaddr(&mmdb, sa, &mmdb_error);
     if (mmdb_error != MMDB_SUCCESS || !lookup.found_entry) {
         return "Unknown";
     }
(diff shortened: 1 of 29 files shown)


src/geoip.cpp
```diff
@@ -18,6 +18,40 @@ GeoDB::~GeoDB()
     MMDB_close(&mmdb);
 }
 
+std::string GeoDB::getGeoLocString(const uint32_t ip_address) const {
```
fcelda: I find this interface very confusing because of the endianness.

`sin_addr` within `sockaddr_in` is network byte order. By looking at the `uint32_t` parameter, I would guess that the function expects host byte order.

src/geoip.cpp
```diff
@@ -18,6 +18,40 @@ GeoDB::~GeoDB()
...
+    MMDB_lookup_result_s lookup = MMDB_lookup_sockaddr(&mmdb, (sockaddr *const)&sa, &mmdb_error);
+    if (mmdb_error != MMDB_SUCCESS || !lookup.found_entry) {
+        return "Unknown";
+    }
+
+    return _getGeoLocString(&lookup);
+
+}
+
+std::string GeoDB::getGeoLocString(const uint8_t* ip_address) const {
```
fcelda: In this case, the used data type doesn't enforce boundary checks.

src/geoip.cpp
```diff
@@ -18,6 +18,40 @@ GeoDB::~GeoDB()
...
+
+    int mmdb_error;
+    struct sockaddr_in6 sa;
+    memset(&sa, 0, sizeof(sa));
+    sa.sin6_addr.s6_addr[0] = ip_address[0];
+    sa.sin6_addr.s6_addr[1] = ip_address[1];
+    sa.sin6_addr.s6_addr[2] = ip_address[2];
+    sa.sin6_addr.s6_addr[3] = ip_address[3];
```
fcelda: The code is copying just 4 bytes of 16-byte IPv6.

src/geoip.cpp
```diff
@@ -112,6 +146,41 @@ std::string GeoDB::_getGeoLocString(MMDB_lookup_result_s* lookup) const {
...
+
+std::string GeoDB::getASNString(const uint8_t* ip_address) const {
+
+    int mmdb_error;
+    struct sockaddr_in6 sa;
+    memset(&sa, 0, sizeof(struct sockaddr_in));
+    sa.sin6_addr.s6_addr[0] = ip_address[0];
+    sa.sin6_addr.s6_addr[1] = ip_address[1];
+    sa.sin6_addr.s6_addr[2] = ip_address[2];
+    sa.sin6_addr.s6_addr[3] = ip_address[3];
```
fcelda: Same problem because of copy-paste.

The guts of `getASNString` and `getGeoLocString` are almost the same. Can this be factored into a helper function returning `MMDB_lookup_result_s`?

src/geoip.h
```diff
@@ -12,10 +12,14 @@ class GeoDB
     ~GeoDB();
 
     std::string getGeoLocString(const char *ip_address) const;
+    std::string getGeoLocString(uint32_t ip_address) const;
```
fcelda: I suggest to drop these new overloads and always use `in_addr` or `in6_addr` (or `sockaddr`) because the types are safer than this.

src/main.cpp
```diff
@@ -277,34 +282,38 @@ void openIface(pcpp::PcapLiveDevice *dev, pktvisor::TcpDnsReassembly &tcpReassem
...
+            pcpp::internal::sockaddr2string(i.addr, buf1);
+            auto nmcvt = pcpp::internal::sockaddr2in6_addr(i.netmask);
             if (!nmcvt) {
                 std::cerr << "couldn't parse IPv4 netmask address on device" << std::endl;
                 continue;
             }
-            hostIPv6.emplace_back(pktvisor::IPv6subnet(pcpp::IPv6Address(adrcvt), pcpp::IPv6Address(nmcvt)));
+            uint8_t len = 0;
+            for (int i = 0; i < 16; i++) {
+                while (nmcvt->s6_addr[i]) {
```
fcelda, line 311: There is `std::popcount` which would do what this loop does. :slightly_smiling_face:

src/querypairmgr.h
```diff
@@ -21,8 +21,8 @@ struct hash_pair {
 };
 
 struct DnsTransaction {
-    timeval queryTS;
-    timeval totalTS;
+    timespec queryTS;
```
fcelda, line 24: I'm wondering about using `std::chrono::time_point` instead. Measurements of the durations and various conversion would come for free from the stdlib. But timespec is fine as well.

src/utils.cpp
```diff
@@ -2,12 +2,13 @@
 #include "utils.h"
 #include <cstring>
 #include <sstream>
-#include <IpUtils.h>
 #ifdef __linux__
```
fcelda: I think this ifdef is not needed. `netinet/in.h` is present on Linux.

src/utils.h
```diff
@@ -8,7 +8,7 @@ namespace pktvisor {
 
 // list of subnets we count as "host" to determine direction of packets
 typedef std::pair<pcpp::IPv4Address, pcpp::IPv4Address> IPv4subnet;
-typedef std::pair<pcpp::IPv6Address, pcpp::IPv6Address> IPv6subnet;
+typedef std::pair<pcpp::IPv6Address, uint8_t> IPv6subnet;
```
fcelda: Using actual struct instead of `std::pair` might make the code more readable.

```cpp
struct IPv6subnet {
    pcpp::IPvAddress address;
    uint8_t mask;
}
```

fcelda (approved):
I left a few suggestions and probably found one bug. It looks good to me overall.

src/geoip.cpp
weyrick: Yeah interesting. It's only done this way to pair with:
```
		/**
		 * Converts the IPv4 address into a 4B integer
		 * @return a 4B integer in network byte order representing the IPv4 address
		 */
		inline uint32_t toInt() const;
...
	private:
		uint8_t m_Bytes[4];
	}; // class IPv4Address


	// Implementation of inline methods

	uint32_t IPv4Address::toInt() const
	{
		uint32_t addr;
		memcpy(&addr, m_Bytes, sizeof(m_Bytes));
		return addr;
	}
```

src/geoip.cpp
weyrick: Agreed I'm not happy with this, and here it is meant to pair with
```		/**
		 * Returns a pointer to 16-byte array representing the IPv6 address
		 */
		const uint8_t* toBytes() const { return m_Bytes; }
...
		uint8_t m_Bytes[16];
```

src/geoip.cpp
weyrick: Nice catch, yes I thought that this macro pointed to the 4 byte (long) version of the address, not the byte. But I see on both OSX and linux it points to the byte (and thank god they are the same)
I will fix that, and also add a test. I was lulled into thinking this worked properly because the existing test passed.

src/geoip.h
weyrick: Yes the only reason I moved away from it was because the IPAddress interface changed - it means I would have to build those myself now, but that's possible.

src/main.cpp
weyrick: Awesome I was wondering, I will take a look.

src/querypairmgr.h
weyrick: Yes in early versions I using `std::chrono`, and I wound up switching to measuring with the packet time stamp instead of with the clock - because the former is available for free. One of the upstream changes was to switch from `timeval` to `timespec`, and this just follows that interface change.

src/main.cpp
weyrick: that seems to require C++20

PR #19, opened 2020-11-24
Adding prometheus exporter

BetaFold3:
This should add prometheus exporter for pktvisor


pktvisor_exporter/pktvisor_exporter_test.go
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2016, 2017 Nicolas Lamirault <nicolas.lamirault@gmail.com>
```
weyrick: assuming this is copy paste, please fix

weyrick (changes requested):
Thanks for this! Since there may be other exporters, can you please rename it from `pktvisor_exporter` to `pktvisor_prometheus` or similar? And then please put it under the `reporting/` directory for now, where we can start to organize tools and documentation for central collection and visibility.

pktvisor_exporter/pktvisor_exporter_test.go
tnguyenns1: @weyrick All fixed.

BetaFold3: @weyrick All fixed.

weyrick (approved):

PR #111, opened 2021-07-14
run-pktvisor.sh IFS fix for args with spaces

rlahnemann:
When including a bpf filter with spaces in PKTVISORD_ARGS, the container fails to start. For example, `-b port 53 and port 53000` will demonstrate this.

The default internal field separator(IFS) is parsing the whitespace in the filter and passing the results as arguments to pktvisord. 

One solution to address this problem is to change the field separator to a semicolon as shown in this PR.


weyrick (changes requested):
Thanks for this, I wish I would have realized we needed this for the arguments that require spaces. I hate to break backward compatibility here though. How about we do it differently without breaking existing installations: let's use a new variable that is semi colon delimited, call it `PKTVISOR_ARGS` (without the D). We can change the docs to make that the main variable people should use now, and say that `PKTVISORD_ARGS` is deprecated. But in the script we will still accept it as is (with a warning?). Does that work?

weyrick:
... and then when 3.3.x lands, we will instead recommend that config files be used instead of arguments.

rlahnemann:
How about PKTVISORD_ARGS accepts both? We check for presence of `;`  in the string and change IFS accordingly?

weyrick:
> How about PKTVISORD_ARGS accepts both? We check for presence of `;` in the string and change IFS accordingly?

If that won't results in a false positive then yeah that's a good idea. I guess I can't think of a situation were semi colon should appear otherwise.

weyrick (approved):
great!

PR #114, opened 2021-08-25
Update issue templates

iatdaitan:


weyrick (changes requested):
@iatdaitan thanks but I think this template is too specific to a UI project, can we get something more appropriate to a backend server?

PR #116, opened 2021-09-01
address corrade ci failure

weyrick:
pin to working corrade recipe


weyrick:
this is due to breaking change in corrade recipe, notified upstream here https://github.com/helmesjo/conan-magnum/issues/6#issuecomment-911697335

PR #156, opened 2021-12-06
implement dynamic modules

weyrick:
add Dynamic Module support. adds `--module-list`, `--module-load`, and `--module-dir` command line options, plus a sample dynamic Mock Handler module.


src/CoreRegistry.cpp
```diff
@@ -12,22 +12,33 @@
...
+CoreRegistry::CoreRegistry()
 {
 
     _logger = spdlog::get("visor");
     if (!_logger) {
         _logger = spdlog::stderr_color_mt("visor");
     }
 
+    // inputs
+    _input_manager = std::make_unique<InputStreamManager>();
```
ghost, line 24: Only a question: is this module dynamically loaded? If so, I didn't get how it is created before it os loaded @weyrick

weyrick: no, this one is static and part of core code

ghost: Oh ok, nice

ghost (approved):

PR #169, opened 2022-01-06
Feature/166 module handlers sequence

### LINKED ISSUE #166: RFC 76 update: accept both a map and an array of handlers at "modules" key, implement chaining
Currently [RFC 76](https://github.com/ns1labs/pktvisor/blob/develop/RFCs/2021-04-16-76-collection-policies.md) specifies a map of handlers attached to an input module. Since a map is unordered by nature, all of the handlers will execute against the input.

We need a way to instead specify a chain/pipeline of handlers, whereby handlers further down in the chain may not execute if handlers further up in the chain decided that the processing should stop (based on a filter). This means that instead of all the handlers connecting to the input stream, the handlers will connect to each other. The handlers have to have explicit support for this to work, and the policy will fail to apply if they are not compatible.

Explicit example: It should be possible to chain PCAP -> DNS -> Net, where Net is able to connect to the DNS module and only process L3/L4 network metrics if DNS re-exposes a signal which indicates that DNS processed and was not filtered. This will allow us to capture, for example, Network metrics only for certain QNAMES that we're specified in the DNS module.

This ticket is complete when:
- [x] RFC 76 documents both map and array versions of `modules` section
- [x] PolicyManager can parse both map and array versions of `modules` section (with unit tests)
- [x] When a chain is present, PolicyManager instantiates the handlers and tries to connect them to each other in a chain instead of all of them to the input handler, failing gracefully if the modules will not connect properly (with unit tests)
- [x] DNS module exposes a signal that Net can connect to, and calls it if not filtered
- [x] Net module is able to connect to DNS module

ghost:


cmd/pktvisord/main.cpp
```diff
@@ -143,10 +144,12 @@ void fill_cmd_options(std::map<std::string, docopt::value> args, CmdOptions &opt
             config_file = YAML::LoadFile(args["--config"].asString());
 
             if (!config_file.IsMap() || !config_file["visor"]) {
-                throw std::runtime_error("invalid schema");
+                logger->error("invalid schema");
```
weyrick: let's be clearer here with "invalid schema in config file: {}", args["--config"].asString()

cmd/pktvisord/main.cpp
```diff
@@ -143,10 +144,12 @@ void fill_cmd_options(std::map<std::string, docopt::value> args, CmdOptions &opt
...
+                exit(EXIT_FAILURE);
             }
             if (!config_file["version"] || !config_file["version"].IsScalar() || config_file["version"].as<std::string>() != "1.0") {
-                throw std::runtime_error("missing or unsupported version");
+                logger->error("missing or unsupported version");
```
weyrick: "... in config file: {}", args["--config"].asString()

src/handlers/dhcp/DhcpStreamHandler.cpp
```diff
@@ -6,9 +6,13 @@
 
 namespace visor::handler::dhcp {
 
-DhcpStreamHandler::DhcpStreamHandler(const std::string &name, InputStream *stream, const Configurable *window_config)
+DhcpStreamHandler::DhcpStreamHandler(const std::string &name, InputStream *stream, const Configurable *window_config, StreamHandler *handler)
     : visor::StreamMetricsHandler<DhcpMetricsManager>(name, window_config)
 {
+    if (handler) {
+        throw StreamHandlerException(fmt::format("DhcpStreamHandler: unsupported stream handler {}", handler->name()));
```
weyrick: "unsupported upstream chained stream handler {}"

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -134,6 +138,7 @@ void DnsStreamHandler::process_udp_packet_cb(pcpp::Packet &payload, PacketDirect
         DnsLayer dnsLayer(udpLayer, &payload);
         if (!_filtering(dnsLayer, dir, l3, pcpp::UDP, metric_port, stamp)) {
             _metrics->process_dns_layer(dnsLayer, dir, l3, pcpp::UDP, flowkey, metric_port, stamp);
+            udp_signal(payload, dir, l3, flowkey, stamp);
```
weyrick, line 142: let's add a comment here 
```
// signal for chained stream handlers, if we have any
```

weyrick (changes requested):
nice! just a few minor changes for clarity.

weyrick (approved):

PR #172, opened 2022-01-07
Remove stopped InputStream and StreamHandlers from Registry managers

### LINKED ISSUE #170: segfault after trying to include a previous deleted policy
pktvisor is generating segfault and also error if you delete an policy and try to add the same deleted again.

Logs
```
[2022-01-07 10:09:52.112] [visor] [info] policy [default]: stopping
[2022-01-07 10:09:52.113] [visor] [info] policy [default]: stopping input instance: default-default
[2022-01-07 10:09:53.894] [visor] [info] policy [default]: stopping handler instance: default-net
[2022-01-07 10:09:53.894] [visor] [info] policy [default]: stopping handler instance: default-dhcp
[2022-01-07 10:09:53.894] [visor] [info] policy [default]: stopping handler instance: default-dns
[2022-01-07 10:09:53.894] [visor] [info] policy [default]: stopping handler instance: default-pcap_stats
[2022-01-07 10:09:53.894] [visor] [info] REQUEST: DELETE /api/v1/policies/default 200
[2022-01-07 10:10:05.365] [visor] [info] policy [default]: parsing
[2022-01-07 10:10:05.365] [visor] [info] policy [default]: instantiating Tap: default
[2022-01-07 10:10:05.366] [visor] [info] policy [default]: instantiating Handler net of type net
[2022-01-07 10:10:05.376] [visor] [info] policy [default]: instantiating Handler dhcp of type dhcp
[2022-01-07 10:10:05.383] [visor] [info] policy [default]: instantiating Handler dns of type dns
[2022-01-07 10:10:05.405] [visor] [info] policy [default]: instantiating Handler pcap_stats of type pcap
[2022-01-07 10:10:05.408] [visor] [info] policy [default]: starting
[2022-01-07 10:10:05.408] [visor] [info] policy [default]: starting input instance: default-default
[2022-01-07 10:10:05.450] [visor] [info] policy [default]: starting handler instance: default-net
[2022-01-07 10:10:05.450] [visor] [info] policy [default]: starting handler instance: default-dhcp
[2022-01-07 10:10:05.450] [visor] [info] policy [default]: starting handler instance: default-dns
[2022-01-07 10:10:05.450] [visor] [info] policy [default]: starting handler instance: default-pcap_stats
[2022-01-07 10:10:05.450] [visor] [info] REQUEST: POST /api/v1/policies 500
[2022-01-07 10:10:05.451] [visor] [error] {"error":"policy [default] creation failed (input): module name 'default-default' already exists"}
Segmentation fault
```

ghost:


weyrick (changes requested):
As discussed, let's encapsulate this into PolicyManager

src/Policies.cpp
```diff
@@ -276,6 +276,28 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
+
+    auto policy = _map[name].get();
+    std::string input_name = policy->input_stream()->name();
+    std::vector<std::string> module_names;
+    for (auto &mod : policy->modules()) {
+        module_names.push_back(mod->name());
+    }
+    policy->stop();
+    _map.erase(name);
+    lock.unlock();
```
ghost: @weyrick I've applied the same previous logic here to avoid locking HandlerManger and InputManager mutexes inside PolicyManager. here we could erase from map only at the end and keep everything in PolicyManger lock context. What do you think is safer?

src/Policies.cpp
```diff
@@ -276,6 +276,28 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
 
+void PolicyManager::remove_policy(const std::string &name)
+{
+    std::unique_lock lock(_map_mutex);
+    if (_map.count(name) == 0) {
+        throw ModuleException(name, fmt::format("module name '{}' does not exist", name));
+    }
+
+    auto policy = _map[name].get();
+    std::string input_name = policy->input_stream()->name();
```
weyrick: tiny but if you use `auto` here, we'll get a const reference instead of a copy here. good habit i think.

src/Policies.cpp
```diff
@@ -276,6 +276,28 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
+    std::vector<std::string> module_names;
```
weyrick, line 288: this can be const ref too maybe?

src/Policies.cpp
```diff
@@ -276,6 +276,28 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
+    for (auto &mod : policy->modules()) {
```
weyrick: const auto&

src/Policies.cpp
```diff
@@ -276,6 +276,28 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
+        module_names.push_back(mod->name());
+    }
+    policy->stop();
+    _map.erase(name);
+    lock.unlock();
+
+    for (auto &name : module_names) {
```
weyrick: const auto&

src/Policies.cpp
weyrick: Yes let's unlock at the end - otherwise we may have a race if the policy is recreated quickly and the handler or input names collide

weyrick (changes requested):

src/Policies.cpp
ghost: The component type of containers like vectors must be assignable

weyrick (approved):

weyrick (approved):

PR #179, opened 2022-01-17
Improve dynamic module load

### LINKED ISSUE #165: Improve Dynamic Plugins load
There are some issues related to dynamic plugin:
1. Dynamic Plugins are being listed as Input and Handler even it has only one type (Input or Handler). Which will fail when instantiate is called.
 a. The Mock handler (dynamic) currently works because static input mock has same alias and it is being loaded instead of mock handler (see item 3 for more info)

2. If user sets a module folder with multiple modules. Even if user does not load all of them. Corrade will list them on pluginList and aliasList and CoreRegistry will try to initialize them and it will fail.

3. Static mock input module and Handler dynamic module has the same alias "mock" which is a problem for Corrade and we should add preferredplugin logic or don't allow two modules with same alias: 
```
        /**
         * @brief Set preferred plugins for given alias
         *
         * By default, if more than one plugin provides given @p alias, one of
         * them is arbitrarily chosen. With this function it's possible to
         * control the behavior. For given @p alias the function goes through
         * the list in @p plugins and uses the first plugin that is available.
         * The @p alias is expected to exist and be defined by plugins in
         * @p plugins. If none of @p plugins is available, nothing is done.
         *
         * Note that after calling @ref setPluginDirectory() or @ref reloadPluginDirectory()
         * this preference gets reset and you may need to call this function
         * again.
         */
        void setPreferredPlugins(const std::string& alias, std::initializer_list<std::string> plugins);
```

ghost:
- Perform load module logic in CoreRegistry
   - add type to all static and dynamic modules
   - filter plugin list by type

### DIFF
--- a/cmd/pktvisord/main.cpp
+++ b/cmd/pktvisord/main.cpp
@@ -70,7 +70,6 @@ static const char USAGE[] =
                                   Please see https://pktvisor.dev for more information
     Modules:
       --module-list               List all modules which have been loaded (builtin and dynamic)
-      --module-load FILE          Load the specified dynamic module
       --module-dir DIR            Set module search path
     Logging Options:
       --log-file FILE             Log to the given output file name
@@ -125,7 +124,6 @@ struct CmdOptions {
 
     struct Module {
         bool list{false};
-        std::pair<bool, std::string> load{false, ""};
         std::pair<bool, std::string> dir{false, ""};
     };
     Module module;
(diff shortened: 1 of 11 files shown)


cmd/pktvisord/main.cpp
```diff
@@ -70,7 +70,6 @@ static const char USAGE[] =
                                   Please see https://pktvisor.dev for more information
     Modules:
       --module-list               List all modules which have been loaded (builtin and dynamic)
-      --module-load FILE          Load the specified dynamic module
       --module-dir DIR            Set module search path
```
weyrick: Change description to "Set module load path. All modules in this directory will be loaded."

weyrick (changes requested):

weyrick (approved):

PR #183, opened 2022-01-19
Unique name for TopN metrics

### LINKED ISSUE #158: prometheus "name" label in top K metric should be unique per metric
Right now all TopK metrics render their prometheus output with a "name" label containing the frequent item:

```
dns_top_qname3{instance="gw",name=".fe.apple-dns.net"} 2
dns_top_qname3{instance="gw",name="clientservices.googleapis.com"} 2
dns_top_qname3{instance="gw",name="gs.apple.com"} 2
dns_top_qname3{instance="gw",name="cdn2.ghostery.com"} 2
dns_top_qtype{instance="gw",name="A"} 30
dns_top_qtype{instance="gw",name="HTTPS"} 8
dns_top_qtype{instance="gw",name="AAAA"} 2
```

This makes "name" very high cardinality, and also hampers data exploration and graphing in e.g. Grafana, which is unable the create an automatic filter on "name" because it is ambiguous which name should be filtered.

We should change this label to be specific per TopK metric, for example "qtype3" and "qtype" in the example above. This will shard the cardinality out, and give us a way to take advantage of Grafana "Filter by Value"

HOWEVER this is a **breaking change** and will break existing dashboards that already use "name" so we need to handle that appropriately.

ghost:


src/Metrics.h
```diff
@@ -234,11 +234,13 @@ class TopN final : public Metric
 private:
     datasketches::frequent_items_sketch<T> _fi;
     size_t _top_count = 10;
+    std::string _schema_name;
```
weyrick: too many things called "schema", can you make this `_item_key` please

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -393,8 +393,16 @@ void DnsMetricsBucket::to_json(json &j) const
     _dns_slowXactOut.to_json(j);
 
     _dns_topUDPPort.to_json(j, [](const uint16_t &val) { return std::to_string(val); });
-    _dns_topQname2.to_json(j);
-    _dns_topQname3.to_json(j);
+    _dns_topQname2.to_json(j, [](std::string val) {
```
weyrick: so we'll want to do this during aggregation instead of output https://github.com/ns1labs/pktvisor/blob/ecd9f67fab805123e07f5ac0d40a610f3c0a0b35/src/handlers/dns/DnsStreamHandler.cpp#L562

otherwise the top N datasketch aggregation will still count differently cased names separately, which I think isn't what we want here

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -93,16 +93,16 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
...
-        , _dns_topREFUSED("dns", {"top_refused"}, "Top QNAMES with result code REFUSED")
-        , _dns_topSRVFAIL("dns", {"top_srvfail"}, "Top QNAMES with result code SRVFAIL")
-        , _dns_topUDPPort("dns", {"top_udp_ports"}, "Top UDP source port on the query side of a transaction")
-        , _dns_topQType("dns", {"top_qtype"}, "Top query types")
-        , _dns_topRCode("dns", {"top_rcode"}, "Top result codes")
-        , _dns_slowXactIn("dns", {"xact", "in", "top_slow"}, "Top QNAMES in transactions where host is the server and transaction speed is slower than p90")
-        , _dns_slowXactOut("dns", {"xact", "out", "top_slow"}, "Top QNAMES in transactions where host is the client and transaction speed is slower than p90")
+        , _dns_topQname2("dns", "qname", {"top_qname2"}, "Top QNAMES, aggregated at a depth of two labels")
+        , _dns_topQname3("dns", "qname", {"top_qname3"}, "Top QNAMES, aggregated at a depth of three labels")
+        , _dns_topNX("dns", "nxdomain", {"top_nxdomain"}, "Top QNAMES with result code NXDOMAIN")
```
weyrick: qname

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -93,16 +93,16 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _dns_topREFUSED("dns", "refused", {"top_refused"}, "Top QNAMES with result code REFUSED")
```
weyrick: qname

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -93,16 +93,16 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _dns_topSRVFAIL("dns", "srvfail", {"top_srvfail"}, "Top QNAMES with result code SRVFAIL")
```
weyrick: qname

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -93,16 +93,16 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _dns_topUDPPort("dns", "udp_ports", {"top_udp_ports"}, "Top UDP source port on the query side of a transaction")
```
weyrick: port

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -93,16 +93,16 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _dns_topQType("dns", "qtype", {"top_qtype"}, "Top query types")
+        , _dns_topRCode("dns", "rcode", {"top_rcode"}, "Top result codes")
+        , _dns_slowXactIn("dns", "xact", {"xact", "in", "top_slow"}, "Top QNAMES in transactions where host is the server and transaction speed is slower than p90")
```
weyrick: qname

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -93,16 +93,16 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _dns_slowXactOut("dns", "xact", {"xact", "out", "top_slow"}, "Top QNAMES in transactions where host is the client and transaction speed is slower than p90")
```
weyrick: qname

weyrick (changes requested):

src/handlers/dns/DnsStreamHandler.cpp
ghost: Right

weyrick (approved):

weyrick (approved):

ghost:
This PR changes the aggregated qname dns to lower case. This changes the qname cardinality as well.

PR #186, opened 2022-01-21
Use same input if for policy if they have same tap and config

### LINKED ISSUE #181: efficient use of taps (and their input streams) which have the same filter

ghost:
- Only create Input Stream if the config and tap is different
    - generate hash from Configurable data
    - Add vector of policies that are using the input in InputStream
    - InputStream name : tap + "-" + hash of input config

### DIFF
--- a/src/Configurable.h
+++ b/src/Configurable.h
@@ -11,6 +11,7 @@
 #include <nlohmann/json.hpp>
 #include <regex>
 #include <shared_mutex>
+#include <sstream>
 #include <string>
 #include <unordered_map>
 #include <variant>
@@ -162,12 +163,13 @@ public:
         }
     }
 
-    void config_hash(std::string &hash)
+    const std::string config_hash()
     {
         std::shared_lock lock(_config_mutex);
-        std::string data;
+        std::vector<std::string> key_values;
         for (const auto &[key, value] : _config) {
-            std::visit([&data, key = key](auto &&arg) {
+            std::visit([&key_values, key = key](auto &&arg) {
+                std::string data;
                 data += key;
                 using T = std::decay_t<decltype(arg)>;
                 if constexpr (std::is_same_v<T, StringList>) {
@@ -179,12 +181,19 @@ public:
                 } else {
                     data += std::to_string(arg);
                 }
+                key_values.push_back(data);
             },
                 value);
         }
-        std::sort(data.begin(), data.end());
-        auto h1 = std::hash<std::string>{}(data);
-        hash = std::to_string(h1);
+        std::sort(key_values.begin(), key_values.end());
+        std::string hash;
+        for (auto &data : key_values) {
+            hash += data;
+        }
+        auto h1 = std::hash<std::string>{}(hash);
+        std::stringstream string_stream;
+        string_stream << std::hex << h1;
+        return string_stream.str();
     }
 };
 
--- a/src/Configurable.h
+++ b/src/Configurable.h
@@ -173,7 +173,9 @@ public:
                 data += key;
                 using T = std::decay_t<decltype(arg)>;
                 if constexpr (std::is_same_v<T, StringList>) {
-                    for (const auto &s : arg) {
+                    auto temp_list = arg;
+                    std::sort(temp_list.begin(), temp_list.end());
+                    for (const auto &s : temp_list) {
                         data += s;
                     }
                 } else if constexpr (std::is_same_v<T, std::string>) {
--- a/src/Configurable.h
+++ b/src/Configurable.h
@@ -161,6 +161,31 @@ public:
             }
         }
     }
+
+    void config_hash(std::string &hash)
+    {
+        std::shared_lock lock(_config_mutex);
+        std::string data;
+        for (const auto &[key, value] : _config) {
+            std::visit([&data, key = key](auto &&arg) {
+                data += key;
+                using T = std::decay_t<decltype(arg)>;
+                if constexpr (std::is_same_v<T, StringList>) {
+                    for (const auto &s : arg) {
+                        data += s;
+                    }
+                } else if constexpr (std::is_same_v<T, std::string>) {
+                    data += arg;
+                } else {
+                    data += std::to_string(arg);
+                }
+            },
+                value);
+        }
+        std::sort(data.begin(), data.end());
+        auto h1 = std::hash<std::string>{}(data);
+        hash = std::to_string(h1);
+    }
 };
 
 class Config : public Configurable
--- a/src/InputStream.h
+++ b/src/InputStream.h
@@ -22,7 +22,7 @@ public:
 
     virtual ~InputStream(){};
 
-    void set_policy(const Policy *policy)
+    void add_policy(const Policy *policy)
     {
         std::unique_lock lock(_input_mutex);
         _policies.push_back(policy);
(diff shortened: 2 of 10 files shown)


src/Configurable.h
```diff
@@ -161,6 +161,31 @@ class Configurable
             }
         }
     }
+
+    void config_hash(std::string &hash)
```
weyrick: modern guidance is to not use "out" parameters, but use a return value instead https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-out

src/Configurable.h
```diff
@@ -161,6 +161,31 @@ class Configurable
...
+                    }
+                } else if constexpr (std::is_same_v<T, std::string>) {
+                    data += arg;
+                } else {
+                    data += std::to_string(arg);
+                }
+            },
+                value);
+        }
+        std::sort(data.begin(), data.end());
```
weyrick: if this is just sorting all of the letters, i think there may be cases where the hash is not unique for some key/value combinations:

a=bc
b=ac

and 

b=bc
a=ac

both yield

aabbcc

even though the values for a and b are different.

If that's the case, I think we need to instead generate a string like this in `data`: "a=acb=bc" (sorted with key/val pairs intact) and hash that instead. Keeping the association solves the problem because we have "a=acb=bc" and "a=bcb=ac" which will hash differently.

src/InputStream.h
```diff
@@ -20,6 +22,24 @@ class InputStream : public AbstractRunnableModule
 
     virtual ~InputStream(){};
 
+    void set_policy(const Policy *policy)
```
weyrick: let's use `add_policy`

src/Configurable.h
```diff
@@ -161,6 +161,31 @@ class Configurable
...
+        auto h1 = std::hash<std::string>{}(data);
+        hash = std::to_string(h1);
```
weyrick: let's convert this to a hex string, which I think will be shorter and read easier in the logs when it becomes part of the name

src/Policies.cpp
```diff
@@ -118,30 +118,40 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
-            // ensure tap input type matches policy input tap
-            if (input_node["input_type"].as<std::string>() != tap->input_plugin()->plugin()) {
-                throw PolicyException(fmt::format("input_type for policy specified tap '{}' doesn't match tap's defined input type: {}/{}", tap_name, input_node["input_type"].as<std::string>(), tap->input_plugin()->plugin()));
+        InputStream *input_ptr;
+        if (_registry->input_manager()->module_exists(input_stream_module_name)) {
+            spdlog::get("visor")->info("policy [{}]: input stream already exists. reusing: {}", policy_name, input_stream_module_name);
+            std::unique_lock<std::shared_mutex> input_lock;
+            auto result_input = _registry->input_manager()->module_get_locked(input_stream_module_name);
+            input_ptr = result_input.module;
+            input_lock = std::move(result_input.lock);
```
weyrick, line 134: this lock will be released immediately because input_lock is defined in this block - I think it needs to move out of the block to where input_stream is defined

src/Policies.cpp
```diff
@@ -299,7 +312,10 @@ void PolicyManager::remove_policy(const std::string &name)
     for (const auto &name : module_names) {
         _registry->handler_manager()->module_remove(name);
     }
-    _registry->input_manager()->module_remove(input_name);
+
+    if (policy->input_stream()->policies_count()) {
```
weyrick: Isn't this the opposite of what we want - shouldn't it be `!policy->input_stream()->policies_count()`? In other words, we only want to remove the module when there are no more policies...

If that's true, there must not be a test case for this, let's make one.

weyrick (changes requested):

src/Policies.cpp
ghost: yes. I missed the "!" here

src/Configurable.h
ghost: @weyrick So you think we can do a vector of strings where each string is "key+values" ? 
then sort the string vector and we concat the vector in one string and hash it

weyrick: yes!

src/Policies.cpp
ghost: lock outside the scope will cause deadlock, probably we should not lock at all. what do you think?

src/Configurable.h
```diff
@@ -161,6 +162,39 @@ class Configurable
...
+    {
+        std::shared_lock lock(_config_mutex);
+        std::vector<std::string> key_values;
+        for (const auto &[key, value] : _config) {
+            std::visit([&key_values, key = key](auto &&arg) {
+                std::string data;
+                data += key;
+                using T = std::decay_t<decltype(arg)>;
+                if constexpr (std::is_same_v<T, StringList>) {
+                    for (const auto &s : arg) {
```
ghost: @weyrick  do you think we should sort StringList here before concat values here? If so, we should create a temp vector because it is const, right?

src/Policies.cpp
weyrick: well, the lock ensures input modules can't change while the lock is held. we need this so that the input module we are trying to use isn't removed and destructed before we make the call to `input_ptr->add_policy(policy_ptr);` at the bottom. what is the deadlock?

weyrick: we definitely need a few test cases with removals

ghost: we do have test cases to cover that. Actually, there is a throw to avoid deadlock in the new test case:
```
/home/lparente/git/pktvisor/src/tests/test_policies.cpp:636: FAILED:
1:   REQUIRE_NOTHROW( registry.policy_manager()->load(config_file2["visor"]["policies"]) )
1: due to unexpected exception with message:
1:   Resource deadlock avoided
```

ghost: When I lock the input mutex in the context

ghost: So, just to clarify. The deadlock I've mentioned was running Unit tests also mentioned above.

I found the race point. The deadlock was that `input_lock` is the mutex from input manager map. Latter on in the Policy Load code there is
```
         if (!_registry->input_manager()->module_exists(input_stream_module_name)) {
                _registry->input_manager()->module_add(std::move(input_stream));
            }
```

module_exists also uses the same mutex to interate over inputStreams map

src/Policies.cpp
```diff
@@ -256,6 +259,10 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
             throw PolicyException(fmt::format("policy [{}] creation failed (policy): {}", policy_name, e.what()));
         }
         try {
+            //Avoid deadlock in input manager
```
ghost: @weyrick  check if you agree with latest changes. moved the policy add to lock context and release the mutex before calling input_manager again

src/Configurable.h
weyrick: good point, yes you are right

src/Policies.cpp
weyrick: Ok this is the right idea, but the problem is we still have a chance of a race between the unlock() and the module_exists/module_add call. I think what we should do here for certainly is make the module_* functions which need a unique_lock take an optional argument where we can move this lock in, and it can use it. That will then guarantee that there will be no races.

weyrick (changes requested):

weyrick (approved):

PR #203, opened 2022-02-06
Cross compilation support

fcelda:


.github/workflows/cross.yml
```diff
@@ -0,0 +1,188 @@
+---
+name: Cross Compilation
+on:
+  push:
```
fcelda, line 4: This needs to be adjusted.

PR #212, opened 2022-02-09
Feature/metrics groups

### LINKED ISSUE #118: implement metric groups
This work will gather metrics from handler modules into groups, allowing them to be turned on or off per policy

ghost:


src/AbstractMetricsManager.h
```diff
@@ -314,13 +314,12 @@ class AbstractMetricsManager
     }
 
 public:
-    AbstractMetricsManager(const Configurable *window_config)
+    AbstractMetricsManager(const Configurable *window_config, const std::bitset<64> groups = std::bitset<64>())
```
ghost: need to add default here because I was getting problem during build because window merge logic

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -21,7 +21,7 @@
 namespace visor::handler::dns {
 
 DnsStreamHandler::DnsStreamHandler(const std::string &name, InputStream *stream, const Configurable *window_config, StreamHandler *handler)
-    : visor::StreamMetricsHandler<DnsMetricsManager>(name, window_config)
+    : StreamMetricsHandler(name, window_config, _process_dns_groups(window_config))
```
ghost: It is a trick here. The first thing I need to do is call parent constructor. So, I cannot initialize group bitset ("_g_enabled") before calling this. so I need to call a function that returns a bitset

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -35,6 +35,8 @@ DnsStreamHandler::DnsStreamHandler(const std::string &name, InputStream *stream,
     if (!_pcap_stream && !_mock_stream && !_dnstap_stream) {
         throw StreamHandlerException(fmt::format("DnsStreamHandler: unsupported input stream {}", stream->name()));
     }
+
+    _g_enabled = _process_dns_groups(window_config);
```
ghost: I tried to use `_g_enabled `inside `_process_dns_groups` so I could initialize it there. however, when `_process_dns_groups ` returns, `_g_enabled` is reinitialized as default, due to c++ initializer order

ghost: So, I need to recall `_process_dns_groups` to initialize `_g_enabled` properly

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -309,6 +313,30 @@ bool DnsStreamHandler::_filtering(DnsLayer &payload, [[maybe_unused]] PacketDire
     return true;
 }
 
+const std::bitset<64> DnsStreamHandler::_process_dns_groups(const Configurable *metrics_config)
```
ghost: I could not return a pointer here because I could initialize the group into a DnsStreamHandler variable before passing it to StreamMetricsHandler.

src/AbstractMetricsManager.h
weyrick: i think we can define a static const integer to 64 and use that everywhere instead of literal "64" in all of the definitions, in case we want to change it later?

weyrick: i think the build problem on window merge logic is legitimate: this line will need to pass in the groups now right?
https://github.com/ns1labs/pktvisor/blob/7aab97ad909ef876c90b507ab13900ba4a09ebc8/src/AbstractMetricsManager.h#L482

so maybe we should remove the defaults on fix all of the calling locations

src/handlers/dns/DnsStreamHandler.cpp
weyrick: Yeah this part is messy, I'd like to avoid the tricks. It seems maybe the best thing to do is to remove group setup from construction and do it at a later phase in the life cycle - perhaps on or just before start() is called by the policy mgr? maybe there's a `configure_groups()` which needs to propagate the group information from the yaml config to the metricsmgr/buckets at that point. Still not great maybe but I like it better then the tricks with the constructors.

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -222,11 +224,13 @@ void DnsStreamHandler::tcp_message_ready_cb(int8_t side, const pcpp::TcpStreamDa
         // data is freed upon return
     };
 
-    if (!iter->second.sessionData[side]) {
-        iter->second.sessionData[side] = std::make_unique<TcpSessionData>(got_dns_message);
-    }
+    if (_g_enabled.test(group::DnsMetrics::TopDnsWire)) {
```
weyrick: this should not be here - this tcp session data is unrelated to groups

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -309,6 +313,30 @@ bool DnsStreamHandler::_filtering(DnsLayer &payload, [[maybe_unused]] PacketDire
     return true;
 }
 
+const std::bitset<64> DnsStreamHandler::_process_dns_groups(const Configurable *metrics_config)
+{
+    // default enabled groups
+    std::bitset<64> groups;
+    groups.set(group::DnsMetrics::Counters);
```
weyrick: we should probably define the list of defaults in the header file - otherwise we will forget to add to the defaults here when we add a new group.

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -309,6 +313,30 @@ bool DnsStreamHandler::_filtering(DnsLayer &payload, [[maybe_unused]] PacketDire
...
+    groups.set(group::DnsMetrics::TopDnsWire);
+    groups.set(group::DnsMetrics::DnsTransactions);
+    groups.set(group::DnsMetrics::TopQnames);
+
+    if (metrics_config->config_exists("enable")) {
+        for (const auto &group : metrics_config->config_get<StringList>("enable")) {
+            groups.set(_group_metrics.at(group));
```
weyrick: this `.at()` call we'll be processing arbitrary user input, so we need to catch when it doesn't match a key and ignore/warn

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -257,7 +273,17 @@ class DnsStreamHandler final : public visor::StreamMetricsHandler<DnsMetricsMana
...
     std::vector<std::string> _f_qnames;
 
+    const std::map<std::string, group::DnsMetrics> _group_metrics = {
+        {"cardinality", group::DnsMetrics::Cardinality},
+        {"counters", group::DnsMetrics::Counters},
+        {"dns_transaction", group::DnsMetrics::DnsTransactions},
+        {"top_dns_wire", group::DnsMetrics::TopDnsWire},
+        {"top_qnames", group::DnsMetrics::TopQnames}};
+
+    std::bitset<64> _g_enabled;
```
weyrick: you do think some of this can exist in the base classes? it would be nice to not have to implement this in every handler.

weyrick (changes requested):

src/handlers/dns/DnsStreamHandler.h
ghost: you mean the enum types? I don't think so because we cannot extend a enum type

weyrick: no I more meant the _g_enabled variable and whatever common methods we can pull out for dealing with initializing/checking groups. that's something we can refactor later though, once we have a better idea of what the implementation really looks like. also, as we discussed, reflection support will affect this work

src/AbstractMetricsManager.h
```diff
@@ -407,6 +415,15 @@ class AbstractMetricsManager
         return _metric_buckets.at(period).get();
     }
 
+    void configure_groups(const std::bitset<GROUP_SIZE> *groups) {
```
ghost: @weyrick  do you think that is right?

src/AbstractMetricsManager.h
```diff
@@ -22,6 +22,8 @@
 
 namespace visor {
 
+constexpr size_t GROUP_SIZE = 64;
```
ghost, line 25: added the group size here because it is the top level header and did as constexpr to ensure that it is available at compile time

src/handlers/dns/DnsStreamHandler.cpp
ghost: I moved this to the StreamHandler class. So I think I could not add default list in header. what do you think?

src/handlers/dns/DnsStreamHandler.cpp
ghost: It is being called in the start, where we also throw other things. So, you think we should simple handle this here and not throw to upstream?

src/handlers/dns/tests/test_dnstap.cpp
```diff
@@ -36,10 +36,10 @@ TEST_CASE("Parse DNSTAP", "[dnstap][dns]")
     CHECK(counters.xacts_in.value() == 0);
     CHECK(counters.xacts_out.value() == 0);
     CHECK(counters.xacts_timed_out.value() == 0);
-    CHECK(counters.NOERROR.value() == 0);
+    CHECK(counters.NOERROR.value() == 70);
```
ghost, line 39: we were not counting response error code before.

src/handlers/dns/DnsStreamHandler.cpp
weyrick: well I am mostly worried that the built in exception isn't useful enough and we want a custom error message - for example, to show the accepted group names

src/StreamHandler.h
```diff
@@ -38,12 +38,35 @@ class StreamHandler : public AbstractRunnableModule
...
 protected:
     std::unique_ptr<MetricsManagerClass> _metrics;
+    std::bitset<GROUP_SIZE> _groups;
+
+    void process_groups(const GroupDefType &group_defs)
+    {
+
+        if (config_exists("enable")) {
+            for (const auto &group : config_get<StringList>("enable")) {
+                _groups.set(group_defs.at(group));
```
weyrick: so I think we want the extra error handling here to include the available group names (keys from group_defs)

src/AbstractMetricsManager.h
```diff
@@ -180,6 +184,7 @@ class AbstractMetricsBucket
                 _end_tstamp.tv_sec = other._end_tstamp.tv_sec;
             }
             _rate_events.merge(other._rate_events);
+            _groups = other._groups;
```
weyrick, line 188: i guess this should actually `|=` to combine the groups

src/AbstractMetricsManager.h
```diff
@@ -195,6 +200,10 @@ class AbstractMetricsBucket
         }
     }
 
+    void configure_groups(const std::bitset<GROUP_SIZE> *groups) {
+        _groups = groups;
```
weyrick, line 207: i think technically this should be protected with `std::unique_lock w_lock(_base_mutex);`

src/AbstractMetricsManager.h
weyrick: yes

src/Policies.cpp
```diff
@@ -231,9 +232,26 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
                     throw PolicyException(fmt::format("invalid stream handler config for handler '{}': {}", handler_module_name, e.what()));
                 }
             }
+            Config handler_metrics;
+            if (module["metrics"]) {
```
weyrick: I'm realizing this is confusing - let's change this (in RFC and code) to "metric_groups"

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -26,6 +26,26 @@ using namespace visor::input::pcap;
 using namespace visor::input::dnstap;
 using namespace visor::input::mock;
 
+// DNS Groups
+namespace group {
+enum DnsMetrics : visor::MetricGroupIntType {
+    Cardinality,
+    Counters,
+    DnsTransactions,
+    TopDnsWire,
```
weyrick: this is unused, let's remove

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -383,45 +397,55 @@ void DnsMetricsBucket::to_json(json &j) const
...
-    _counters.replies.to_json(j);
-    _counters.TCP.to_json(j);
-    _counters.UDP.to_json(j);
-    _counters.IPv4.to_json(j);
-    _counters.IPv6.to_json(j);
-    _counters.NX.to_json(j);
-    _counters.REFUSED.to_json(j);
-    _counters.SRVFAIL.to_json(j);
-    _counters.NOERROR.to_json(j);
+    if ((*_groups)[group::DnsMetrics::Counters]) {
```
weyrick: this is really awkward, how about we make a method in AbstractMetricsBucket:
```
    inline bool group_enabled(int32_t g)
    {
        return (*_groups)[g];
    }
```
and use that instead

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -383,45 +397,55 @@ void DnsMetricsBucket::to_json(json &j) const
...
+        _counters.UDP.to_json(j);
+        _counters.IPv4.to_json(j);
+        _counters.IPv6.to_json(j);
+        _counters.NX.to_json(j);
+        _counters.REFUSED.to_json(j);
+        _counters.SRVFAIL.to_json(j);
+        _counters.NOERROR.to_json(j);
 
-    _counters.filtered.to_json(j);
+        _counters.filtered.to_json(j);
```
weyrick: even though it is a counter, this one should not be part of this group i think

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -436,29 +460,17 @@ void DnsMetricsBucket::process_dnstap(bool deep, const dnstap::Dnstap &payload)
 {
     std::unique_lock lock(_mutex);
 
+    pcpp::ProtocolType l3;
     if (payload.message().has_socket_family()) {
         if (payload.message().socket_family() == dnstap::INET6) {
-            ++_counters.IPv6;
+            l3 = pcpp::IPv6;
```
weyrick, line 471: it looks like you removed these counters because we are potentially counting them in `process_dns_layer` - but that will only be called if the `query_message` or `response_message` is provided by dnstap, which it may not be - and if it's not, it should do the counting here. so I think we need to keep this counting code if there is no query or response message

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -26,6 +26,26 @@ using namespace visor::input::pcap;
...
+    TopQnames
+};
+}
+
+enum Protocol : uint64_t {
+    DNSTAP_UDP = 0x01,
```
weyrick: please document exactly where these two groups come from (dnstap.proto and pcpp/ProtocolType.h) because it's important that these do not overlap - and both may change outside of our control as those projects progress.

in fact, instead of defining them using the literals, can we define them from the values in those headers?

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -26,6 +26,26 @@ using namespace visor::input::pcap;
...
+    DNSTAP_TCP = 0x02,
+    DNSTAP_DOT = 0x03,
+    DNSTAP_DOH = 0x04,
+    PCAP_TCP = 0x08,
```
weyrick: please make this prefix `PCPP_` instead of `PCAP_`

weyrick (changes requested):

src/AbstractMetricsManager.h
ghost: the bucket stores only a pointer to the handler bitset, i cannot perform this operation here.

src/StreamHandler.h
ghost: done

src/AbstractMetricsManager.h
ghost: ok, added that

src/handlers/dns/DnsStreamHandler.cpp
ghost: actually we have two process_dns_layer. but your affirmation was right in case of deep sampling and no query/response message. I've added an else there now to process  this correctly

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -481,79 +494,90 @@ void DnsMetricsBucket::process_dnstap(bool deep, const dnstap::Dnstap &payload)
...
-    if (payload.message().has_query_port()) {
-        _dns_topUDPPort.update(payload.message().query_port());
-    }
-
     if (payload.message().has_query_zone()) {
         // TODO decode wire name, use in top_qname
     }
 
     if (!deep || (!payload.message().has_query_message() && !payload.message().has_response_message())) {
+        process_dns_layer(l3, l4, side, 0);
```
ghost, line 508: @weyrick  actually it was already right the logic here. don't need the extra else that i've added

src/Policies.cpp
ghost: done

src/handlers/dns/DnsStreamHandler.cpp
ghost: done

src/handlers/dns/DnsStreamHandler.h
ghost: done

src/handlers/dns/DnsStreamHandler.h
ghost: i've used changed that to the values in those headers

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -385,45 +392,55 @@ void DnsMetricsBucket::to_json(json &j) const
...
-        }
-    });
+
+    if (group_enabled(group::DnsMetrics::TopQnames)) {
+        _dns_topQname2.to_json(j);
+        _dns_topQname3.to_json(j);
+        _dns_topNX.to_json(j);
+        _dns_topREFUSED.to_json(j);
+        _dns_topSRVFAIL.to_json(j);
+        _dns_topRCode.to_json(j, [](const uint16_t &val) {
```
weyrick: topRCode is not in group

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -655,46 +713,55 @@ void DnsMetricsBucket::to_prometheus(std::stringstream &out, Metric::LabelMap ad
...
-        }
-    });
+
+    if (group_enabled(group::DnsMetrics::TopQnames)) {
+        _dns_topQname2.to_prometheus(out, add_labels);
+        _dns_topQname3.to_prometheus(out, add_labels);
+        _dns_topNX.to_prometheus(out, add_labels);
+        _dns_topREFUSED.to_prometheus(out, add_labels);
+        _dns_topSRVFAIL.to_prometheus(out, add_labels);
+        _dns_topRCode.to_prometheus(out, add_labels, [](const uint16_t &val) {
```
weyrick: topRCode not in group

weyrick (changes requested):

weyrick (approved):

PR #216, opened 2022-02-14
Create dnstap_msg_type filter in DNS handler

### LINKED ISSUE #204: Dns filter for DNSTAP input
Allow filtering dnstap signal in DNS module by the Message_Type protobuf:
```
enum Message_Type : int {
  Message_Type_AUTH_QUERY = 1,
  Message_Type_AUTH_RESPONSE = 2,
  Message_Type_RESOLVER_QUERY = 3,
  Message_Type_RESOLVER_RESPONSE = 4,
  Message_Type_CLIENT_QUERY = 5,
  Message_Type_CLIENT_RESPONSE = 6,
  Message_Type_FORWARDER_QUERY = 7,
  Message_Type_FORWARDER_RESPONSE = 8,
  Message_Type_STUB_QUERY = 9,
  Message_Type_STUB_RESPONSE = 10,
  Message_Type_TOOL_QUERY = 11,
  Message_Type_TOOL_RESPONSE = 12,
  Message_Type_UPDATE_QUERY = 13,
  Message_Type_UPDATE_RESPONSE = 14
};

```

As a dnstap user, I want to be able to create policies which extract DNS metrics per "type" defined above: Authoritative, Resolver, Client, Forwarder, Stub, Tool or Update.

Therefor DNS handler needs a new filter `dnstap_msg_type` which can be set to one of "auth", "resolver", "client", "forwarder", "stub", "tool" or "update" to filter out all dnstap messages except those which match the given type (which will match both QUERY and RESPONSE of that type).

ghost:


src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -73,6 +73,17 @@ void DnsStreamHandler::start()
...
     }
+    if (config_exists("dnstap_msg_type")) {
+        auto type = config_get<std::string>("dnstap_msg_type");
+        try {
+            auto type_pair = _dnstap_map_types.at(type);
+            _f_dnstap_types.set(type_pair.first);
+            _f_dnstap_types.set(type_pair.second);
+            _f_enabled.set(Filters::DnstapMsgType);
+        } catch (const std::exception &e) {
+            throw ConfigException(fmt::format("dnstap_msg_type error {}", e.what()));
```
weyrick, line 84: let's go ahead and dump the keys in _dnstap_map_types in this message so the user knows which they can set

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -743,6 +760,9 @@ void DnsMetricsManager::process_dnstap(const dnstap::Dnstap &payload)
     // base event
     new_event(stamp);
     // process in the "live" bucket. this will parse the resources if we are deep sampling
+    if (filtered) {
+        live_bucket()->process_filtered();
```
weyrick, line 764: shouldn't this have a return after it, or the `process_dnstap` below be in an else { } ? i'm confused why the unit test passes if so.

src/handlers/dns/DnsStreamHandler.cpp
ghost: ok

src/handlers/dns/DnsStreamHandler.cpp
ghost: I've called livebucket here that only performs 
```
void DnsMetricsBucket::process_filtered()
{
    std::unique_lock lock(_mutex);
    ++_counters.filtered;
}
```

should I change the logic here to call metricsManager process_filtered and return?

weyrick: right calling `process_filtered` is good, but then it goes on to call `process_dnstap` with the payload, which means it would actually count all those metrics and not really filter it right?

PR #217, opened 2022-02-16
Feature/filter dnstap by hostmask

### LINKED ISSUE #205: dnstap filter by host mask
As a dnstap user, I want to be able to extract DNS metrics only for dnstap signals matching a specific list of IP netmasks (IPv4 and IPv6). Because dnstap is not able to use BPF filters like PCAP, the filtering must be done in user space in the dnstap module.

Therefore, dnstap input needs a new filter `only_hosts` (string array) which may contain hostmasks in CIDR form:
```
"192.168.0.0/24",
"1.2.3.4/32",
"2001:db8::/64"
```

dnstap input will convert these to IPv4 and IPv6 IP+netmasks and filter out any dnstap message whose `query_address` or `response_address` does not match a hostmask in the list.

This functionality is very similar to Pcap input handler's "host_spec" functionality, and that maybe used as a reference. See https://github.com/ns1labs/pktvisor/blob/7aab97ad909ef876c90b507ab13900ba4a09ebc8/src/inputs/pcap/utils.cpp#L68 and https://github.com/ns1labs/pktvisor/blob/7aab97ad909ef876c90b507ab13900ba4a09ebc8/src/inputs/pcap/PcapInputStream.cpp#L365

ghost:


src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -323,6 +379,81 @@ void DnstapInputStream::_create_frame_stream_unix_socket()
...
+                throw DnstapException(fmt::format("invalid IPv4 address: {}", ip));
+            }
+            _IPv4_host_list.emplace_back(ipv4, cidr_number);
+        }
+    }
+}
+
+bool DnstapInputStream::_match_subnet(const std::string &dnstap_ip)
+{
+    if (dnstap_ip.length() == 16) {
```
ghost: could not test ipv6 match filter.

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -323,6 +379,85 @@ void DnstapInputStream::_create_frame_stream_unix_socket()
...
+                throw DnstapException(fmt::format("invalid IPv4 address: {}", ip));
+            }
+            _IPv4_host_list.emplace_back(ipv4, cidr_number);
+        }
+    }
+}
+
+bool DnstapInputStream::_match_subnet(const std::string &dnstap_ip)
+{
+    if (dnstap_ip.size() == 16) {
```
ghost: could not test ipv6 match filter. we don't have a dnstap file with this input

src/inputs/dnstap/DnstapInputStream.h
```diff
@@ -11,19 +11,15 @@
...
 #include <unordered_map>
-#include <uv.h>
-
-namespace uvw {
-class Loop;
-class AsyncHandle;
-class PipeHandle;
-class TCPHandle;
-}
+#include <uvw.hpp>
```
weyrick: curious why this change?

src/inputs/dnstap/DnstapInputStream.h
```diff
@@ -40,10 +36,16 @@ class DnstapInputStream : public visor::InputStream
     std::shared_ptr<uvw::TCPHandle> _tcp_server_h;
     std::unordered_map<uv_os_fd_t, std::unique_ptr<FrameSessionData<uvw::TCPHandle>>> _tcp_sessions;
 
+    std::vector<Ipv4Subnet> _IPv4_host_list;
+    std::vector<Ipv6Subnet> _IPv6_host_list;
+    bool _filter_host;
```
weyrick: we probably will have more filters here - shouldn't we just use the bitset?

src/inputs/dnstap/DnstapInputStream.cpp
weyrick: can we do a unit and just call this manually with a valid ipv6 bytestring?

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -323,6 +379,85 @@ void DnstapInputStream::_create_frame_stream_unix_socket()
...
+            if ((result || prefixLength < 8) && compareBitCount > 0) {
+                uint8_t subSubnetByte = network.s6_addr[compareByteCount] >> (8 - compareBitCount);
+                uint8_t subThisByte = ipv6.s6_addr[compareByteCount] >> (8 - compareBitCount);
+                result = subSubnetByte == subThisByte;
+            }
+            if (result) {
+                return true;
+            }
+        }
+    } else if (dnstap_ip.size() == 4) {
```
weyrick: i wonder if it's worth checking _IPv4_host_list.size() here too, to avoid memcpy when there's no IPv4 in the list (same for ipv6 above)

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -60,6 +56,24 @@ void DnstapInputStream::_read_frame_stream_file()
                 _logger->warn("dnstap data is wrong type or has no message, skipping frame of size {}", len_data);
                 continue;
             }
+            if (_filter_host) {
+                if (d.message().has_query_address()) {
```
weyrick: hm i think this logic needs some work. i believe dnstap will most often provide both the query _and_ the response address, and it should pass the filter if it matches either. if dnstap provides only one or the other, then it needs to just match that one i guess.

weyrick (changes requested):

src/inputs/dnstap/DnstapInputStream.cpp
ghost: agree

src/inputs/dnstap/DnstapInputStream.h
ghost: We are using uvw which is a c++ wrapper over libuv. By including the wrapper header intead of libuv we dont need to declare the classes that exists on uvw (TCPHandle, etc).

weyrick: right I did this originally as a compile time optimization though - uvw includes a lot of things we don't need in the interface definition here. so the idea is that the forward declarations are lighter weight and speed things up.

ghost: Oh sorry. I just remove my changes in this part.

src/inputs/dnstap/DnstapInputStream.cpp
ghost: changed the logic. please evaluate again.

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -56,15 +60,14 @@ void DnstapInputStream::_read_frame_stream_file()
                 _logger->warn("dnstap data is wrong type or has no message, skipping frame of size {}", len_data);
                 continue;
             }
-            if (_filter_host) {
-                if (d.message().has_query_address()) {
-                    if (!_match_subnet(d.message().query_address())) {
+            if (_f_enabled[Filters::OnlyHosts]) {
+                if (d.message().has_query_address() && d.message().has_response_address()) {
+                    if (!_match_subnet(d.message().query_address()) || !_match_subnet(d.message().response_address())) {
```
weyrick: this `||` should be `&&` because with `||` it means "skip if either doesn't match" but with `&&` it means "skip if both don't match" (i.e. at least one needs to match) and we want the latter. double check that but i'm pretty sure it's right. to be clear: it's ok to have query OR response OR both match, but at least one has to or else we filter.

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -56,15 +60,14 @@ void DnstapInputStream::_read_frame_stream_file()
...
                         continue;
                     }
-                    if (d.message().has_response_address()) {
-                        if (!_match_subnet(d.message().response_address())) {
-                            continue;
-                        }
+                } else if (d.message().has_query_address()) {
```
weyrick: this (and below) can collapse to one if now, makes it a bit easier to read I think

weyrick (changes requested):

src/inputs/dnstap/DnstapInputStream.cpp
ghost: Oh ok, now i got it

src/inputs/dnstap/DnstapInputStream.cpp
ghost: collapse in a `||` comparison? because I should know if it exists to perform match;

weyrick: i just meant 
```
} else if (d.message().has_query_address() && !_match_subnet(d.message().query_address())) {
    continue;
}
```

ghost: ok

ghost: done

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -60,6 +60,19 @@ void DnstapInputStream::_read_frame_stream_file()
...
+                if (d.message().has_query_address() && d.message().has_response_address()) {
+                    if (!_match_subnet(d.message().query_address()) && !_match_subnet(d.message().response_address())) {
+                        continue;
+                    }
+                } else if (d.message().has_query_address() && !_match_subnet(d.message().query_address())) {
+                    continue;
+                } else if (d.message().has_response_address() && !_match_subnet(d.message().response_address())) {
+                    continue;
+                } else {
+                    continue;
```
weyrick: please doc `// message had neither query nor response address, so filter`

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -60,6 +60,19 @@ void DnstapInputStream::_read_frame_stream_file()
                 _logger->warn("dnstap data is wrong type or has no message, skipping frame of size {}", len_data);
                 continue;
             }
+            if (_f_enabled[Filters::OnlyHosts]) {
+                if (d.message().has_query_address() && d.message().has_response_address()) {
+                    if (!_match_subnet(d.message().query_address()) && !_match_subnet(d.message().response_address())) {
+                        continue;
```
weyrick: please doc `// message had both query and response address, and neither matched, so filter`

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -60,6 +60,19 @@ void DnstapInputStream::_read_frame_stream_file()
...
+                    }
+                } else if (d.message().has_query_address() && !_match_subnet(d.message().query_address())) {
+                    continue;
```
weyrick: please doc `// message had only query address and it didn't match, so filter`

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -60,6 +60,19 @@ void DnstapInputStream::_read_frame_stream_file()
...
+                } else if (d.message().has_response_address() && !_match_subnet(d.message().response_address())) {
+                    continue;
```
weyrick: please doc `// message had only response address and it didn't match, so filter`

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -60,6 +60,19 @@ void DnstapInputStream::_read_frame_stream_file()
                 _logger->warn("dnstap data is wrong type or has no message, skipping frame of size {}", len_data);
                 continue;
             }
+            if (_f_enabled[Filters::OnlyHosts]) {
```
weyrick: let's break this code out into a method which excepts a const ref to the message and returns a bool, and use it in file/udp/tcp instead of repeating these blocks.

weyrick (changes requested):

src/inputs/dnstap/DnstapInputStream.cpp
ghost: I think it is not good to throw things in critical path. what about only returning true or false?

weyrick (approved):

weyrick (approved):

PR #225, opened 2022-03-07
Only process pcap raw packets if it is possible

ghost:


src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -319,6 +319,10 @@ void PcapInputStream::_generate_mock_traffic()
 
 void PcapInputStream::process_raw_packet(pcpp::RawPacket *rawPacket)
 {
+    if (processing_packet) {
+        return;
```
ghost, line 323: we should probably add a counter here to count not processed packeges

src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -319,6 +319,10 @@ void PcapInputStream::_generate_mock_traffic()
 
 void PcapInputStream::process_raw_packet(pcpp::RawPacket *rawPacket)
 {
+    if (processing_packet) {
```
leoparente, line 322: used a atomic bool in this case because it will have less overhead then a mutex lock and try lock

src/inputs/pcap/PcapInputStream.cpp
ghost: I've used a atomic bool in this case because it will have less overhead then a mutex lock and try lock

ghost:
not possible solution

PR #229, opened 2022-03-10
Feature/cache handler data

ghost:


src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -849,4 +917,40 @@ void DnsMetricsManager::process_dnstap(const dnstap::Dnstap &payload, bool filte
...
+void DnsMetricsManager::process_cache_pcap(CacheDnsHandler *cache, PacketDirection dir, uint32_t flowkey)
+{
+    new_event(cache->timestamp);
+    if (cache->payload.getHeaderLen() == 0) {
+        live_bucket()->process_dns_layer(cache->l3, cache->l4, cache->side, cache->port, false);
+
+    } else {
+        live_bucket()->process_dns_layer(_deep_sampling_now, cache->payload, cache->l3, cache->l4, cache->port, false);
+
+        if (group_enabled(group::DnsMetrics::DnsTransactions)) {
```
ghost, line 937: DNS is not caching transaction information. should we figure out a way to do that?

PR #230, opened 2022-03-10
Implement ThreadMonitor for linux systems

ghost:


src/ThreadMonitor.h
```diff
@@ -0,0 +1,72 @@
...
+#if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
+        return 0;
+#elif __APPLE__
+        return 0;
+#elif __linux__
+        uint64_t memory;
+        std::string token;
+        std::ifstream file("/proc/thread-self/status");
+        while (file >> token) {
+            if (token == "VmRSS:") {
```
ghost: should we get only VmRSS for memory or should we use another approach?

src/ThreadMonitor.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include <fstream>
+#include <vector>
+
+namespace visor {
+
+class ThreadMonitor
+{
+public:
+    static inline double cpu_percentage()
```
ghost: this is not right yet. I would like to know which approach we should use to get cpu percentage. most of them uses a time delay and calculate the clock difference between this time

src/handlers/resources/ResourcesStreamHandler.cpp
```diff
@@ -0,0 +1,157 @@
...
+void ResourcesStreamHandler::start()
+{
+    if (_running) {
+        return;
+    }
+
+    if (config_exists("recorded_stream")) {
+        _metrics->set_recorded_stream();
+    }
+
```
ghost: The idea is to attach to the input callbacks that should be alway called to ensure that resource usage will be called sometimes

ghost: User will decide when they want to attach this handler to their policy or not

src/handlers/resources/ResourcesStreamHandler.cpp
```diff
@@ -0,0 +1,157 @@
...
+    } else if (_sflow_stream) {
+        _sflow_connection.disconnect();
+    }
+
+    _running = false;
+}
+
+void ResourcesStreamHandler::process_sflow_cb([[maybe_unused]] const SFSample &)
+{
+    if (difftime(time(NULL), _timer) >= MEASURE_INTERVAL) {
```
ghost: I used time_t to calculate diff time to get resources in the fixed interval (10 seconds)
the reason to use time_t is that we don't need precision here and it is faster then clocks methods:
https://gist.github.com/stunpix/54c304bc010307562ec3434dbd5e76ae

src/Policies.cpp
```diff
@@ -174,6 +174,18 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
         }
 
+        std::unique_ptr<Policy> input_resources_policy;
+        std::unique_ptr<StreamHandler> resources_module;
+        if (input_stream) {
+            // create new policy with resources handler for input stream
+            input_resources_policy = std::make_unique<Policy>(input_stream_module_name + "-resources", tap);
+            input_resources_policy->set_input_stream(input_ptr);
+            auto resources_handler_plugin = _registry->handler_plugins().find("input_resources");
+            resources_module = resources_handler_plugin->second->instantiate(input_stream_module_name + "-resources", input_ptr, &window_config);
```
weyrick: let's put a check to make sure resources_handler_plugin was found, just in case

src/Policies.cpp
```diff
@@ -174,6 +174,18 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
             window_config.config_set<uint64_t>("deep_sample_rate", _default_deep_sample_rate);
         }
 
+        std::unique_ptr<Policy> input_resources_policy;
+        std::unique_ptr<StreamHandler> resources_module;
+        if (input_stream) {
```
weyrick, line 183: since the input_stream may be being reused here, it needs to check if the input_resources_policy has already been added to this input_stream, right? otherwise we'd get > 1

src/Policies.cpp
```diff
@@ -268,6 +280,9 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
 
         // make sure policy starts before committing
         try {
+            if (input_resources_policy) {
+                input_resources_policy->start();
```
weyrick, line 291: this should only start() it if it hadn't been already (although maybe we handle that in start() already)

src/Policies.cpp
```diff
@@ -336,8 +355,13 @@ void PolicyManager::remove_policy(const std::string &name)
         _registry->handler_manager()->module_remove(name);
     }
 
-    if (!policy->input_stream()->policies_count()) {
+    if (policy->input_stream()->policies_count() <= 1) {
```
weyrick: let's not make any assumptions here - if there's only 1 policy left, get it and make sure it's the input_resources policy before trying to clean it up. we may make adding the input_resources policy optional, for example

weyrick (changes requested):

src/Policies.cpp
ghost: This input_stream is the unique pointer. it only exists when new input is create, otherwise we will only have the input_ptr.

src/Policies.cpp
ghost: input_resources_policy unique ptr will only exists if new input and therefore new input_resources_policy exists. it will be called only once

src/Policies.cpp
ghost: I've done this validation in another way, please check it

src/Policies.cpp
```diff
@@ -278,12 +295,16 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
         } catch (ModuleException &e) {
             throw PolicyException(fmt::format("policy [{}] creation failed (policy): {}", policy_name, e.what()));
         }
         try {
-            if (input_stream) {
+            if (input_stream && resources_module) {
                 _registry->input_manager()->module_add(std::move(input_stream));
+                _registry->handler_manager()->module_add(std::move(resources_module));
             }
         } catch (ModuleException &e) {
```
weyrick, line 312: this and the other catch ModuleException needs to possibly rollback the resources handler

src/Policies.cpp
```diff
@@ -336,7 +357,16 @@ void PolicyManager::remove_policy(const std::string &name)
         _registry->handler_manager()->module_remove(name);
     }
 
-    if (!policy->input_stream()->policies_count()) {
+    if (policy->input_stream()->policies_count() == 1) {
```
weyrick: ok i think this is ok, but let's comment the strategy here, i.e. "if there is only one policy left on the input stream, and that policy is the input resources policy, then remove it"

src/Policies.cpp
```diff
@@ -336,7 +357,16 @@ void PolicyManager::remove_policy(const std::string &name)
...
+        auto input_resources_name = input_name + "-resources";
+        if (_map.count(input_resources_name) != 0) {
+            auto resources_policy = _map[input_resources_name].get();
+            resources_policy->stop();
+            _registry->handler_manager()->module_remove(input_resources_name);
+            _registry->input_manager()->module_remove(input_name);
+            _map.erase(input_resources_name);
+        }
+    } else if (!policy->input_stream()->policies_count()) {
```
weyrick: this shouldn't be "else if", just "if" right, because if we just removed the input resources policy, this will be 0 and we want to remove the input

weyrick (changes requested):

src/Policies.cpp
ghost: I mean I was removing in both == 1 and == 0, but changed the logic now

src/Policies.cpp
ghost: ok

src/Policies.cpp
ghost: agreed. I've improved the rollback logic

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -33,14 +33,15 @@ class InputResourcesMetricsBucket final : public visor::AbstractMetricsBucket
     // total numPackets is tracked in base class num_events
     Quantile<double> _cpu_percentage;
     Quantile<uint64_t> _memory_usage_kb;
```
weyrick: this says "_kb" but the metric says "bytes". let's keep it consistent with metric name and make it "_memory_bytes"

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -0,0 +1,123 @@
...
+    // total numPackets is tracked in base class num_events
+    Quantile<double> _cpu_percentage;
+    Quantile<uint64_t> _memory_usage_kb;
+    Counter _policies_number;
+    Counter _handlers_count;
+    bool _merged;
+
+public:
+    InputResourcesMetricsBucket()
+        : _cpu_percentage("resources", {"cpu_percentage"}, "Quantiles of thread cpu usage")
```
weyrick: let's make this "cpu_usage", both the metric name and our variables for consistency. description "Quantiles of 5s averages of CPU usage by the input stream"

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -0,0 +1,123 @@
...
+
+constexpr double MEASURE_INTERVAL = 10; // in seconds
+
+class InputResourcesMetricsBucket final : public visor::AbstractMetricsBucket
+{
+
+protected:
+    mutable std::shared_mutex _mutex;
+
+    // total numPackets is tracked in base class num_events
```
weyrick: remove old comment

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -0,0 +1,123 @@
...
+    Quantile<double> _cpu_percentage;
+    Quantile<uint64_t> _memory_usage_kb;
+    Counter _policies_number;
```
weyrick: let's use _policy_count

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -0,0 +1,123 @@
...
+    Counter _handlers_count;
```
weyrick: _handler_count (singular)

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -0,0 +1,123 @@
...
+    bool _merged;
+
+public:
+    InputResourcesMetricsBucket()
+        : _cpu_percentage("resources", {"cpu_percentage"}, "Quantiles of thread cpu usage")
+        , _memory_usage_kb("resources", {"memory_bytes"}, "Quantiles of thread memory usage in bytes")
```
weyrick: Quantiles of 5s averages of memory usage (in bytes) by the input stream

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -0,0 +1,123 @@
...
+#include <string>
+
+namespace visor::handler::resources {
+
+using namespace visor::input::pcap;
+using namespace visor::input::dnstap;
+using namespace visor::input::mock;
+using namespace visor::input::sflow;
+
+constexpr double MEASURE_INTERVAL = 10; // in seconds
```
weyrick: let's go 5

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -0,0 +1,123 @@
...
+    Quantile<uint64_t> _memory_usage_kb;
+    Counter _policies_number;
+    Counter _handlers_count;
+    bool _merged;
+
+public:
+    InputResourcesMetricsBucket()
+        : _cpu_percentage("resources", {"cpu_percentage"}, "Quantiles of thread cpu usage")
+        , _memory_usage_kb("resources", {"memory_bytes"}, "Quantiles of thread memory usage in bytes")
+        , _policies_number("resources", {"policies_attached"}, "Total number of policies attached to the input stream")
```
weyrick: let's stick with "policy_count"

src/handlers/input_resources/InputResourcesStreamHandler.h
```diff
@@ -0,0 +1,123 @@
...
+        , _handlers_count("resources", {"handlers_attached"}, "Total number of handlers attached to the input stream")
```
weyrick: handler_count

weyrick (changes requested):

weyrick (approved):
🎉

PR #231, opened 2022-03-11
Update 3rd party libs

ghost:
- Update conan lib version to latest available
    - Update datasketches to version 3.3.0


3rd/datasketches/common/CMakeLists.txt
```diff
@@ -29,17 +29,18 @@ target_include_directories(common
...
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/MurmurHash3.h
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/serde.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/count_zeros.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/inv_pow2_table.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/binomial_bounds.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/conditional_back_inserter.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/conditional_forward.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/ceiling_power_of_2.hpp
-)
+install(TARGETS common EXPORT ${PROJECT_NAME})
```
weyrick: Why this change?

ghost: I've just copied and pasted the folders. They Changed this Cmake file. if you look the previous version 2.1.0 and the newer version 3.3.0

weyrick: Yeah ok, I wasn't doing a full copy - this CMake file is handcrafted and needs to stay the same, no install. Also, we don't need the tests

ghost: ok. I will do that for all datasketches folders. I saw also that build fail for ARM due to UVW update (and libuv). I will rollback  that to previous version

weyrick (approved):

PR #233, opened 2022-03-14
Remove not closed PCAP TCP connections

ghost:


src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -245,6 +245,16 @@ void PcapInputStream::tcp_connection_end(const pcpp::ConnectionData &connectionD
 void PcapInputStream::process_pcap_stats(const pcpp::IPcapDevice::PcapStats &stats)
 {
     pcap_stats_signal(stats);
+
+    for (const auto &packet : _tcp_reassembly.getConnectionInformation()) {
+        if (_tcp_reassembly.isConnectionOpen(packet.second)) {
+            time_t start_time = static_cast<time_t>(packet.second.startTime.tv_sec);
+            // check if started is 10 times closeDelayConnection (seconds)
+            if (difftime(time(NULL), start_time) >= 10) {
```
ghost, line 253: we could make  configurable TCP timeout

weyrick: yeah seems reasonable. does this fix it? we don't really want to timeout on the start time, but rather the last time we saw a packet from them - is that available?

ghost: https://github.com/seladb/PcapPlusPlus/blob/d8d702acc99bf8bbf1531f00a3162c6b11a3cd2c/Packet%2B%2B/src/TcpReassembly.cpp#L147

based on the logic from line 147 to line 153 we could use the endTime

ghost: but I think we also need to check if it is not empty.

because it seems that end time is not set for the first packet that arrives

leoparente: actually just by creating a timeval obj it will get the current time right?

ghost: I tried this strategy of getting the endTime and packets never closes. 

based on that, It keep receiving packets of current connections and receiving new connections

ghost: Actually, it seems that info connection map is a copy of the start of the connection only:
https://github.com/seladb/PcapPlusPlus/blob/d8d702acc99bf8bbf1531f00a3162c6b11a3cd2c/Packet%2B%2B/src/TcpReassembly.cpp#L131

So in this case the endTime will always be 0...

weyrick: Do you think this is an upstream bug?

ghost: yes. I've opened an issue for PcapPlusPlus https://github.com/seladb/PcapPlusPlus/issues/843
I will close this PR for now

Issue #232, opened 2022-03-14
DNS stats should be based upon public suffix


mccarthy-james:
DNS stats are calculated in buckets of `Top QName 2` and `Top QName 3`, but this does not work with second-level domain names like .co.uk. It would be better if these stats are calculated based upon the public suffix list (so queries to www.google.co.uk would count towards stats as www.google.co.uk and google.co.uk and not google.co.uk and .co.uk).

Please see https://publicsuffix.org/list/

PR #242, opened 2022-03-28
Feature/adjust qname aggregation

ghost:


ghost:
TEST_CASE("DNS Filters: only_qname_suffix", "[pcap][dns]")


before static suffix:
```
   "top_qname2":[
      {
         "estimate":6,
         "name":".google.com"
      },
      {
         "estimate":4,
         "name":"google.com"
      }
   ],
   "top_qname3":[
      {
         "estimate":4,
         "name":"play.google.com"
      },
      {
         "estimate":2,
         "name":"nonexistent.google.com"
      }
   ]
```

After static suffix:
```
   "top_qname2":[
      {
         "estimate":4,
         "name":"play.google.com"
      },
      {
         "estimate":4,
         "name":"google.com"
      },
      {
         "estimate":2,
         "name":"nonexistent.google.com"
      }
   ],
   "top_qname3":[
      
   ]
```

src/handlers/dns/dns.cpp
```diff
@@ -18,7 +18,9 @@ AggDomainResult aggregateDomain(const std::string &domain)
         return AggDomainResult(qname2, qname3);
     }
     std::size_t endDot = std::string::npos;
-    if (domain.back() == '.') {
+    if (!static_suffix.empty() && domain.size() > static_suffix.size()) {
```
weyrick: if this doesn't need to compare the suffix (because it's already compared I guess?), why not just pass a size_t instead?

src/handlers/dns/tests/test_dns.cpp
```diff
@@ -72,4 +72,31 @@ TEST_CASE("DNS Utilities", "[dns]")
...
         CHECK(result.second == "");
     }
+
+    SECTION("aggregateDomain with static suffix")
+    {
+        AggDomainResult result;
+        std::string domain;
+
+        domain = "biz.foo.bar.com";
+        result = aggregateDomain(domain, ".bar.com");
```
weyrick: this is unintuitive i think - if we're passing a string, I would expect that string to match against domain in some way, so let's change it to a size_t

weyrick (changes requested):

src/handlers/dns/dns.cpp
ghost: it is already compared in the filtering. so I think we should not compare it again. It is better to use size_t only

weyrick:
also, may make sense to consider if this work can help with #232

src/handlers/dns/tests/test_dns.cpp
```diff
@@ -72,4 +72,47 @@ TEST_CASE("DNS Utilities", "[dns]")
...
+        CHECK(result.first == ".bar.com.");
+        CHECK(result.second == "foo.bar.com.");
+
+        domain = "www.google.co.uk";
+        static_suffix = ".co.uk";
+        result = aggregateDomain(domain, static_suffix.size());
+        CHECK(result.first == ".google.co.uk");
+        CHECK(result.second == "www.google.co.uk");
+
+        domain = "www.google.co.uk";
```
ghost, line 112: @weyrick is this the expected result as well?

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -788,12 +791,12 @@ void DnsMetricsBucket::process_filtered()
 }
 
 // the general metrics manager entry point (both UDP and TCP)
-void DnsMetricsManager::process_dns_layer(DnsLayer &payload, PacketDirection dir, pcpp::ProtocolType l3, pcpp::ProtocolType l4, uint32_t flowkey, uint16_t port, timespec stamp)
+void DnsMetricsManager::process_dns_layer(DnsLayer &payload, PacketDirection dir, pcpp::ProtocolType l3, pcpp::ProtocolType l4, uint32_t flowkey, uint16_t port, size_t suffix_size, timespec stamp)
```
weyrick, line 794: note for future work: we should start thinking about moving these variables to a structure

weyrick (approved):

leoparente:
@weyrick   I don't know why the tests are failing on macOS. Based on test results it seems that the suffix size is always 0. it is not getting changed here https://github.com/ns1labs/pktvisor/pull/242/files#diff-a1a3caffbb070ff5e9e83eabc1228e90fd159d63a8363ede0b98525ab0a5b818R327

can you check that locally on you machine when you have some time?

PR #243, opened 2022-03-28
local_thread cached data

ghost:


src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -20,6 +20,8 @@
 
 namespace visor::handler::dns {
 
+thread_local static DnsCacheData cached_dns_layer;
```
leoparente, line 23: @weyrick  thread_local cannot be a member variable:
https://stackoverflow.com/a/10999179

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -241,6 +241,13 @@ class DnsStreamHandler final : public visor::StreamMetricsHandler<DnsMetricsMana
 {
     static constexpr size_t DNSTAP_TYPE_SIZE = 15;
 
+    struct DnsCacheData {
+        uint32_t flowKey = 0;
```
weyrick, line 245: i don't think we should use flowKey at all - after consideration, it's possible to receive two different packets with the same flowKey that need should not reuse cache. timestamp should be more accurate and sufficient

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -241,6 +241,13 @@ class DnsStreamHandler final : public visor::StreamMetricsHandler<DnsMetricsMana
...
+        timespec timestamp = timespec();
+        std::unique_ptr<DnsLayer> dnsLayer;
+    };
+    static thread_local DnsCacheData cached_dns_layer;
```
weyrick: should be `_cached_dns_layer` now that it's a member variable

weyrick (changes requested):

src/handlers/dns/DnsStreamHandler.h
ghost: As I mentioned. using only packet timestamp still not unique enough. IT happens to have packets with same timestamp at least in pktvisor tests. So I need to use both FlowKey AND timestamp (sec and nsec)

ghost: only timestamp breaks pktvisor tests: i.e. get different results from what currently is

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -161,9 +163,14 @@ void DnsStreamHandler::process_udp_packet_cb(pcpp::Packet &payload, PacketDirect
         metric_port = dst_port;
     }
     if (metric_port) {
-        DnsLayer dnsLayer(udpLayer, &payload);
-        if (!_filtering(dnsLayer, dir, l3, pcpp::UDP, metric_port, stamp)) {
-            _metrics->process_dns_layer(dnsLayer, dir, l3, pcpp::UDP, flowkey, metric_port, stamp);
+        if (flowkey != cached_dns_layer.flowKey || stamp.tv_sec != cached_dns_layer.timestamp.tv_sec || stamp.tv_nsec != cached_dns_layer.timestamp.tv_nsec) {
```
ghost: @weyrick  using only packet timestamp still not unique enough. IT happens to have packets with same timestamp at least in pktvisor tests. So I need to use both FlowKey AND timestamp (sec and nsec)

src/handlers/dns/DnsStreamHandler.h
weyrick: we should fix those tests then - i don't see how it's possible to legitimately get a different packet with the exact same time stamp. we'll need to dig in - but that's a better thing to fix then to try to use flowKey

ghost:
@weyrick  as I mentioned yesterday. We should cache transformed qname to no do that multiple times. However, the transform operation happens on Bucket and the cache only exists in handler (member variable). should we back to the Global local cache?

ghost:
So, I've tested develop branch over a huge pcap file with the following policy with the command during **~15 seconds**

`/usr/bin/perf record --freq=997 --call-graph dwarf -q /home/lparente/git/pktvisor/cmake-build-debug/bin/pktvisord --config /home/lparente/default.yaml
`

```
version: "1.0"

visor:
  config:
    admin_api: true
  taps:
    default:
      input_type: pcap
      config:
        iface: "mock"
        pcap_source: mock
  policies:
   default:
      kind: collection
      input:
        tap: default
        input_type: pcap
        filter:
          bpf: "udp"
        config:
          pcap_file: "/home/lparente/huge.pcap"
      handlers:
        window_config:
          num_periods: 5
          deep_sample_rate: 100
        modules:
           dns1:
            type: dns
           dns2:
            type: dns
           dns5:
            type: dns
           net1:
            type: net

```
And I've got the following result
develop_performance


Base on that. We could improve the calls of `std::tranform` and `parseResources ` to be called only once.
Also we could use another container intead of `unordered_map`

ghost:
After this PR changes we performed the same test during the same time (**15 seconds** ).
With cached DnsLayer being parsed only once and std::transform lower case being performed inside `parseResources` we got the following result:
local_cache_and_dns_transform


**BEFORE (Develop)**
std::transform + parseResources = 7.03%+3.8% = **10.83%**

**NOW with CACHE**
parseResources (std::transform inside) = **5.06%**


The result is visible: the cycles of `DnsMetricsBucket::process_dns_layer` dropped from **31.9%** to **26.6%**

ghost:
The another improvement would be on DNS Transactions. Just by replacing the` unordered map` container with robin-hood hash map we got an improvement:
cache_with_robin_hood


We can look at `DnsMetricsManager::process_dns_layer`

develop: **61.6%**
cached: **60.1%**
cached+robin-hood: **55.1%**

I've also tested parallel hash map and tsl robin map. but robin-hood got the best result

weyrick (approved):

PR #244, opened 2022-03-29
Add timeout to tcp reassembly connections

ghost:


src/inputs/pcap/PcapInputStream.h
```diff
@@ -61,6 +63,8 @@ class PcapInputStream : public visor::InputStream
 
     pcpp::TcpReassembly _tcp_reassembly;
 
+    timespec _tcp_timeout;
```
weyrick, line 66: how about `_last_tcp_clean_ts`

src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -379,6 +380,16 @@ void PcapInputStream::process_raw_packet(pcpp::RawPacket *rawPacket)
         case pcpp::TcpReassembly::Ignore_Retransimission:
             break;
         }
+
+        // Remove staled connections every TCP_TIMEOUT period (not every packet)
```
weyrick, line 384: either "remove stalled" or "remove stale" :)

src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -379,6 +380,16 @@ void PcapInputStream::process_raw_packet(pcpp::RawPacket *rawPacket)
...
+        if (rawPacket->getPacketTimeStamp().tv_sec > _tcp_timeout.tv_sec + TCP_TIMEOUT) {
+            _tcp_timeout = rawPacket->getPacketTimeStamp();
+            for (const auto &connection : _tcp_reassembly.getConnectionInformation()) {
+                if (connection.second.endTime.tv_sec > 0 && _tcp_timeout.tv_sec > connection.second.endTime.tv_sec + TCP_TIMEOUT) {
```
weyrick, line 388: i think we need two variables: one for 'how often to clean tcp sessions' which can be 10s and one for 'how long a TCP session can be stale before closing' which should be more like 30s

weyrick (changes requested):

src/inputs/pcap/PcapInputStream.cpp
seladb: 1. I agree with @weyrick that it's better to have 2 variables
2. I'm not sure how many concurrent connection `pktvisor` is designed to handle, but if there are many this loop will block the entire flow every `TCP_TIMEOUT`. There are a few possible solutions here:
   1. Run this cleanup in a separate thread 
   2. Keep an LRU list of [flow keys + last packet timestamps] and go over the connection list from the "least recently used" - check if the timestamp is old and close the connection, then move to the next "least recently used" and so on. PcapPlusPlus has an implementation of an [LRU list](https://github.com/seladb/PcapPlusPlus/blob/master/Common%2B%2B/header/LRUList.h) but it's limited in size
   3. Don't go over more than X connections in the loop. If there are more than X - continue iterating when the next packet arrives. This will require keeping an iterator of the keys in the connection map so might be a little more complex to implement

weyrick: Thanks for the thoughts @seladb. We actually have this same pattern in our DNS transaction tracking system, which also needs a cleanup. I think whatever solution we use, we can apply it there too. I was also discussing the LRU strategy with @leoparente yesterday.

ghost: Many thanks @seladb!

@weyrick I've tested here and with this new approach without calling `isConnectionOpen` which iterates over the real connection map, it seems that we don't have any problem executing the clear connection logic in another thread. 

But do you think that should we move to LRU List approach? if so, should we use pcap LRU list with limited size or should we create our own with no limit?

weyrick: Doesn't doing it in another thread require mutexes around this code and the `_tcp_reassembly.reassemblePacket(packet);` call? This would create contention per tcp packet if it's waiting for the cleaner thread, but hard to say if that is more or less efficient than blocking in the same thread.

ghost: I mean. I was getting error segfault /bad allocation previously because for every connection in connection data map I was performing:

```
int TcpReassembly::isConnectionOpen(const ConnectionData& connection) const
{
	ConnectionList::const_iterator iter = m_ConnectionList.find(connection.flowKey);
	if (iter != m_ConnectionList.end())
		return iter->second.closed == false;

	return -1;
}
```

So main thread (`_tcp_reassembly.reassemblePacket(packet);`) and other thread were both iterating over `m_ConnectionList` and we got map exception.

But now I'm not doing that anymore. HOWEVER calling `closeConnection(flowKey)` also interates over `m_ConnectionList`. Which still unsafe and would require mutex. So I think it is safer to do everything in same thread.

ghost:
Implemented a Solution based on pcpp [LRUList](https://github.com/ns1labs/pktvisor/pull/249). This PR will be closed

Issue #255, opened 2022-04-04
Not correct CPU percentage returned by input resources policy


ghost:

leoparente:
After some testing we realized that cpu percentage was properly working for pktvisor binary and not for pktvisor docker container.
To get this working on docker container one should pass the tag `--pid=host` on `docker run` command.


```
pid: "host"

Sets the PID mode to the host PID mode. This turns on sharing between container and the host 
operating system the PID address space. Containers launched with this flag can access and manipulate
other containers in the bare-metal machine's namespace and vice versa.
```

PR #256, opened 2022-04-04
Bugfix/calculate cpu percentage based on core

ghost:


leoparente:
see https://github.com/ns1labs/pktvisor/issues/255#issuecomment-1102867515

PR #260, opened 2022-04-08
Feature/support to crashpad trace

ghost:


cmd/pktvisord/main.cpp
```diff
@@ -68,6 +69,10 @@ static const char USAGE[] =
     Configuration:
       --config FILE               Use specified YAML configuration to configure options, Taps, and Collection Policies
                                   Please see https://pktvisor.dev for more information
+    Crashpad:
+      --cp-token TOKEN            Specifies Crashpad information for tracing issues remote
```
weyrick: "Crashpad token for remote crash reporting"

cmd/pktvisord/main.cpp
```diff
@@ -68,6 +69,10 @@ static const char USAGE[] =
...
+      --cp-url URL                Crashpad server url
+      --cp-path PATH              Crashpad binary handler
```
weyrick: Crashpad handler binary

cmd/pktvisord/main.cpp
```diff
@@ -68,6 +69,10 @@ static const char USAGE[] =
     Configuration:
       --config FILE               Use specified YAML configuration to configure options, Taps, and Collection Policies
                                   Please see https://pktvisor.dev for more information
+    Crashpad:
```
weyrick, line 72: so we also need a way for users to turn this off, even in cases where they are using a container that hasn't enabled by default. perhaps `--cp-disable` and/or checking an environment variable

cmd/pktvisord/CrashpadHandler.h
```diff
@@ -0,0 +1,44 @@
...
+    std::map<std::string, std::string> annotations;
+    std::vector<std::string> arguments;
+    CrashpadClient client;
+    bool rc;
+
+    annotations["format"] = "minidump";
+    annotations["product"] = "pktvisor";
+    annotations["database"] = "pktvisor";
+    annotations["version"] = VISOR_VERSION_NUM;
+    annotations["token"] = token;
```
weyrick, line 23: is other information like OS, kernel or anything else sent automatically?

cmd/pktvisord/CMakeLists.txt
```diff
@@ -2,9 +2,13 @@ add_executable(pktvisord main.cpp)
 
 target_link_libraries(pktvisord
         PRIVATE
+        ${CONAN_LIBS_CRASHPAD}
```
weyrick, line 5: so we have to figure out how to make this optional at build time

cmd/pktvisord/CrashpadHandler.h
```diff
@@ -0,0 +1,44 @@
...
+    base::FilePath::StringType db_path("crashpad");
```
weyrick, line 24: is this a file on disk? i see "no-rate-limit" below, but we would need to be careful if this writes to disk, we can't let the disk fill

cmd/pktvisord/main.cpp
```diff
@@ -385,6 +415,18 @@ int main(int argc, char *argv[])
...
 
+    // crashpad support
+    if (options.crashpad_info.token.first || options.crashpad_info.url.first || options.crashpad_info.path.first) {
+        if (options.crashpad_info.token.first && options.crashpad_info.url.first && options.crashpad_info.path.first) {
+            if (!crashpad::start_crashpad_handler(options.crashpad_info.token.second, options.crashpad_info.url.second, options.crashpad_info.path.second)) {
+                logger->error("failed to setup crashpad");
+            }
+        }
+        else {
+            logger->error("missing information to setup crashpad");
```
weyrick: in these error cases, should we exit?

weyrick (changes requested):

cmd/pktvisord/CrashpadHandler.h
ghost: Yes it does
```
uname.machine   amd64
uname.release   null
uname.sysname   Linux
uname.version   5.14.0 -1031-oem #34-Ubuntu SMP Fri Mar 25 08:21:05 UTC 2022 x86_64
....
```

cmd/pktvisord/main.cpp
leoparente: Don't get why should we need to add this disable. if the crashpad configurations are no passed correctly, the crashpad code will not be executed

cmd/pktvisord/CMakeLists.txt
leoparente: the code will be not called if config is not passed

cmd/pktvisord/CrashpadHandler.h
leoparente: We can change that on future commit after we get this properly running

cmd/pktvisord/main.cpp
leoparente: yes, i will add that

cmd/pktvisord/CrashpadHandler.h
leoparente: the rate limit should be a dump per hour

cmd/pktvisord/main.cpp
weyrick: If it's only going into the debug container, then that's okay. But if it's going to be enabled by default in the release container, then people need a way to shut it off.

weyrick (approved):

weyrick (approved):

PR #265, opened 2022-04-16
Add crash handler on debug images

etaques:
This PR changes:
- Added stage just to build binaries
- Save binaries produced as artifacts
- Added crashpad_handler binary on container image
- Added crashpad parameter to pktvisord entrypoint
- Push pktvisord (symbol) to backtrace.io after build
- Added Dockerfile.crashhandler file to not mix with old ones


.github/actions/build-go/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+#
+function build() {
+  echo "========================= Building pktvisor-cli ========================="
+  cp -rf golang/ /src/
+  cp -rf ./version.go /src/pkg/client/version.go
```
weyrick, line 7: please document here that this depends on the cpp build step having been run first to produce version.go

weyrick (approved):

PR #275, opened 2022-04-26
Feature/netflow ipfix support

### LINKED ISSUE #263: Implement NetFlow input handler
We are currently able to ingest flow data via [sFlow](https://github.com/ns1labs/pktvisor/tree/develop/src/inputs/sflow), and we have the same use case for [NetFlow](https://en.wikipedia.org/wiki/NetFlow) data. This issue represents the work required to create an Input Handler for ingesting NetFlow data streams. The implementation will be similar to #106

leoparente:


src/inputs/flow/FlowInput.conf
```diff
@@ -0,0 +1,5 @@
+# Aliases
+provides=flow
```
weyrick, line 2: can we do "providers=sflow" too?

src/inputs/flow/FlowInputStream.cpp
```diff
@@ -0,0 +1,204 @@
...
+        auto flow_type = config_get<std::string>("flow_type");
+        if (flow_type == "sflow") {
+            _flow_type = Type::SFLOW;
+        } else if (flow_type == "netflow") {
+            _flow_type = Type::NETFLOW;
+        } else {
+            throw FlowException(fmt::format("invalid flow type: {}", flow_type));
+        }
+    } else {
+        throw FlowException("flow config must specify flow_type");
```
weyrick: default to "sflow" (with a warning log) for backwards compat?

src/inputs/flow/FlowInputStream.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    if (config_exists("flow_type")) {
+        auto flow_type = config_get<std::string>("flow_type");
+        if (flow_type == "sflow") {
+            _flow_type = Type::SFLOW;
+        } else if (flow_type == "netflow") {
+            _flow_type = Type::NETFLOW;
+        } else {
+            throw FlowException(fmt::format("invalid flow type: {}", flow_type));
```
weyrick: let's list the ones we support here in the error message

weyrick (changes requested):

src/inputs/flow/FlowInput.conf
leoparente: done

weyrick (approved):

PR #280, opened 2022-05-03
Feature/flow handler

### LINKED ISSUE #279: Flow Handler Module
Implement flow handler module to handle custom flow parameters

leoparente:


src/handlers/flow/FlowStreamHandler.h
```diff
@@ -0,0 +1,222 @@
...
+
+    struct topns {
+        TopN<std::string> topSrcIP;
+        TopN<std::string> topDstIP;
+        TopN<uint16_t> topSrcPort;
+        TopN<uint16_t> topDstPort;
+        TopN<uint32_t> topInIfIndex;
+        TopN<uint32_t> topOutIfIndex;
+        topns(std::string metric)
+            : topSrcIP("flows", "ip", {"top_src_ip_" + metric}, "Top source IP addresses by " + metric)
```
weyrick: since this defines a public interface, let's stay very consistent with existing metrics (e.g. listed on https://github.com/ns1labs/pktvisor/wiki/Sample-pktvisor-Output-Data)

please make this "flow" (singular) to stay consistent with others which are singular (dns, net, pcap, etc)

it looks like "ips" should be plural though, as in existing `src_ips_in`

there may be others, please take a look. if there's something ambiguous and we have to decide, let me know.

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -0,0 +1,222 @@
...
+            , topDstIP("flows", "ip", {"top_dst_ip_" + metric}, "Top destination IP addresses by " + metric)
+            , topSrcPort("flows", "port", {"top_src_ports_" + metric}, "Top source ports by " + metric)
+            , topDstPort("flows", "port", {"top_dst_ports_" + metric}, "Top destination ports by " + metric)
+            , topInIfIndex("flows", "index", {"top_in_if_index_" + metric}, "Top input interface indexes  by " + metric)
```
weyrick: extra space before "by"

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -0,0 +1,222 @@
...
+            , topOutIfIndex("flows", "index", {"top_out_if_index_" + metric}, "Top output interface indexes by " + metric)
+        {
+        }
+    };
+
+    topns _topByBytes;
+    topns _topByPackets;
+
+    // total numPackets is tracked in base class num_events
```
weyrick: i think this comment doesn't apply, right? or maybe it's numFlowPackets essentially?

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -0,0 +1,222 @@
...
+    FlowMetricsBucket()
+        : _srcIPCard("flows", {"cardinality", "src_ips_in"}, "Source IP cardinality")
+        , _dstIPCard("flows", {"cardinality", "dst_ips_out"}, "Destination IP cardinality")
+        , _topGeoLoc("flows", "geo_loc", {"top_geoLoc"}, "Top GeoIP locations")
+        , _topASN("flows", "asn", {"top_ASN"}, "Top ASNs by IP")
+        , _topByBytes("bytes")
+        , _topByPackets("packets")
+        , _payload_size("flows", {"payload_size"}, "Quantiles of payload sizes, in bytes")
+        , _rate("flows", {"rates", "pps"}, "Rate of flow packets per second")
+        , _throughput("payload", {"rates", "bps"}, "Rate of flow packets size in bytes per second")
```
weyrick: hm this is size of the flow packet itself, not the inner flow data right? Maybe we can make that more clear.

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -0,0 +1,222 @@
...
+    {
+        set_event_rate_info("flows", {"rates", "pps_total"}, "Rate of all packets (combined ingress and egress) in packets per second");
+        set_num_events_info("flows", {"total"}, "Total packets processed");
```
weyrick: again, probably should be "Total number of flow packets processed" ?

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -0,0 +1,222 @@
...
+    bool _filtering(const FlowData &flow);
+
+public:
+    FlowStreamHandler(const std::string &name, InputStream *stream, const Configurable *window_config, StreamHandler *handler = nullptr);
+    ~FlowStreamHandler() override;
+
+    // visor::AbstractModule
+    std::string schema_key() const override
+    {
+        return "flows";
```
weyrick: can we make this a static const somewhere and reference it, rather than a string?

weyrick (changes requested):
Looking really good!

src/handlers/flow/FlowStreamHandler.h
leoparente: the inner flow data

src/handlers/flow/FlowStreamHandler.h
leoparente: this gets the packet that contains the multiple flow packets inside

src/handlers/flow/FlowStreamHandler.h
leoparente: where should we do that? all the handlers are using in this way. should be inside the handler itself due to dynamic loading plugins

weyrick: yes for internal use in the handler, for example in all the metrics definitions

src/handlers/flow/FlowStreamHandler.h
weyrick: i don't understand still

src/handlers/flow/FlowStreamHandler.h
weyrick: so this is the rate of all _combined_ flow inner data? then it should say "Rate of all combined flows in bytes per second"

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -116,19 +118,21 @@ class FlowMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _dstPortCard("flow", {"cardinality", "dst_ports_out"}, "Destination ports cardinality")
+        , _topGeoLoc("flow", "geo_loc", {"top_geoLoc"}, "Top GeoIP locations")
+        , _topASN("flow", "asn", {"top_ASN"}, "Top ASNs by IP")
         , _topByBytes("bytes")
         , _topByPackets("packets")
-        , _payload_size("flows", {"payload_size"}, "Quantiles of payload sizes, in bytes")
-        , _rate("flows", {"rates", "pps"}, "Rate of flow packets per second")
-        , _throughput("payload", {"rates", "bps"}, "Rate of flow packets size in bytes per second")
+        , _payload_size("flow", {"payload_size"}, "Quantiles of payload sizes, in bytes")
+        , _rate("flow", {"rates", "pps"}, "Rate of inner flow data packets per second")
```
weyrick: if I understand this correctly now, then it should be "Rate of combined flow packets per second"

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -116,19 +118,21 @@ class FlowMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _throughput("payload", {"rates", "bps"}, "Rate of inner flow data packets size in bytes per second")
```
weyrick: "Rate of combined flow bytes per second"

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -116,19 +118,21 @@ class FlowMetricsBucket final : public visor::AbstractMetricsBucket
...
     {
-        set_event_rate_info("flows", {"rates", "pps_total"}, "Rate of all packets (combined ingress and egress) in packets per second");
-        set_num_events_info("flows", {"total"}, "Total packets processed");
-        set_num_sample_info("flows", {"deep_samples"}, "Total packets that were sampled for deep inspection");
+        set_event_rate_info("flow", {"rates", "pps_total"}, "Rate of all packets (combined ingress and egress) in packets per second");
```
weyrick: Take these out for now

weyrick (changes requested):

src/handlers/flow/FlowStreamHandler.h
leoparente: done

src/handlers/flow/FlowStreamHandler.h
leoparente: done

src/handlers/flow/FlowStreamHandler.h
leoparente: done

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -171,6 +171,7 @@ class FlowMetricsManager final : public visor::AbstractMetricsManager<FlowMetric
 
 class FlowStreamHandler final : public visor::StreamMetricsHandler<FlowMetricsManager>
 {
+    static constexpr const char *_flow_schema{"flow"};
```
weyrick: hm well can we use this in the metric definitions too then?

weyrick:
I think this is pretty close, though I know @rboucher-ns1 wanted to review the metric names and definitions too. The question is, should we land this so we can start testing functionality ahead of final sign off on metric schema. I think it's ok to do that in the develop branch, just not after it's been released to main in a published version.

src/handlers/flow/FlowStreamHandler.h
leoparente: we could not the way I did. I've just changed it to use on bucket as well

weyrick (approved):
ok let's do this

PR #282, opened 2022-05-05
fixing scenarios and add mocked iface

manrodrigues:
I'm not considering 'period', 'rates', 'quantiles_us', 'payload_size' metrics


automated_tests/features/pktvisor.feature
```diff
@@ -1,70 +1,86 @@
...
+  Then the pktvisor container status must be <pkt_status>
+    And metrics must be correctly generated
+  Examples:
+    | status_port | role | file_name | pkt_status |
+    | available   | user | dns_ipv6_udp.pcap | running    |
+    | available   | user | dns_ipv6_tcp.pcap | running    |
+    | available   | user | dns_ipv4_udp.pcap | running    |
+    | available   | user | dns_ipv4_tcp.pcap | running    |
+    | available   | user | dhcp-flow.pcap | running    |
+    | available   | user | dns_udp_mixed_rcode.pcap | running    |
```
lpegoraro, line 87: Line at end of file.
Check this two options in your IDE.
Screenshot from 2022-05-04 11-33-22
.

lpegoraro (approved):
Good stuff, I just have one thing to ask, if we could create those json files automatically and in runtime, so they wouldn't be fixed on git and be susceptible to become outdated, would this be better? 
If we could create a function that exports these json files as a utils in the python test?

automated_tests/features/pktvisor.feature
manrodrigues: Not sure if I understand. The files I saved because talking to Léo, we could assume that what pktvisor is generating today as metrics can be assumed to be correct. If we generate every moment and compare, we will not be able to guarantee that the metrics are correct, right? 'Cause I'll always be comparing a thing to itself

lpegoraro (approved):
LGTM

lpegoraro:
Great job!

leoparente (approved):
Nice work

PR #284, opened 2022-05-05
Bugfix/period exception

### LINKED ISSUE #228: Handler PeriodException when getting bucket metrics with no data provided by InputStream
the change [remove "flush timer"](https://github.com/ns1labs/pktvisor/pull/192) fixed issues on bucket because it creates a new timer thread to generate empty buckets.

However, for a policy with multiple handlers, If one handler does not receive inputs and the other do, when we try to get bucket it will generate a exception for this particular handler and REST API will return error (425).

leoparente:


src/AbstractMetricsManager.h
```diff
@@ -440,6 +440,16 @@ class AbstractMetricsManager
         }
     }
 
+    void shift_bucket(timespec stamp)
```
weyrick: please refactor name to `check_period_shift`

src/InputStream.h
```diff
@@ -15,6 +15,9 @@ class InputStream : public AbstractRunnableModule
     mutable std::shared_mutex _input_mutex;
     std::vector<const Policy *> _policies;
 
+protected:
+    static constexpr uint8_t RUNNING_NOTIFY_INTERVAL = 30; // in seconds
```
weyrick: please refactor names from "running" to "heartbeat", so `HEARTBEAT_INTERVAL`, `heartbeat_signal`, etc.

src/inputs/dnstap/DnstapInputStream.cpp
```diff
@@ -224,6 +240,7 @@ void DnstapInputStream::_create_frame_stream_tcp_socket()
 
     // spawn the loop
     _io_thread = std::make_unique<std::thread>([this] {
+        _timer->start(uvw::TimerHandle::Time{1000}, uvw::TimerHandle::Time{RUNNING_NOTIFY_INTERVAL * 1000});
```
weyrick: is there a spot where we need to `stop()` or `close()` or something, when the input is stopped (not just on error event?)

weyrick (changes requested):

src/AbstractMetricsManager.h
leoparente: done

src/InputStream.h
leoparente: done

src/inputs/dnstap/DnstapInputStream.cpp
leoparente: based on their unit tests it seems we dont actually need to stop and close it. but I did it on async event as well

leoparente: done

weyrick (approved):

PR #289, opened 2022-05-27
Feature/input filter

### LINKED ISSUE #285: Input Filter should not necessary generate new input handler
For Input Sources that requires bind to ip/port it would be only one instance available.
Currently, If user try to create policies with different filters, a policy will fail to apply because it will try to bind to a port that is already in use.

To overcome this issue, we should change the way that Handlers attach to InputStreams

leoparente:


src/inputs/flow/FlowInputStream.cpp
```diff
@@ -73,7 +73,10 @@ void FlowInputStream::_read_from_pcap_file()
                 sample.rawSampleLen = udpLayer->getLayerPayloadSize();
                 try {
                     read_sflow_datagram(&sample);
-                    sflow_signal(sample);
+                    std::unique_lock lock(_input_mutex);
+                    for (auto &callback : _callbacks) {
+                        dynamic_cast<FlowInputStreamCallback *>(callback.get())->sflow_cb(sample);
```
leoparente: don't like this approach. but I could not use Template with InputStream Class

src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -374,17 +392,26 @@ void PcapInputStream::process_raw_packet(pcpp::RawPacket *rawPacket)
...
+        dynamic_cast<PcapInputStreamCallback *>(callback.get())->process_packet_cb(packet, dir, l3, l4, timestamp);
+    }
 
     if (l4 == pcpp::UDP) {
-        udp_signal(packet, dir, l3, pcpp::hash5Tuple(&packet), timestamp);
+        for (auto &callback : _callbacks) {
+            dynamic_cast<PcapInputStreamCallback *>(callback.get())->process_udp_packet_cb(packet, dir, l3, pcpp::hash5Tuple(&packet), timestamp);
+        }
     } else if (l4 == pcpp::TCP) {
+        lock.unlock();
```
leoparente, line 405: don't know about this part as well. need to release and lock the mutex again (line 411) to not have deadlock. could change it to recursive_mutex

src/InputStream.h
```diff
@@ -51,20 +101,36 @@ class InputStream : public AbstractRunnableModule
         return _policies.size();
     }
 
-    virtual size_t consumer_count() const
+    size_t consumer_count() const
     {
-        return policy_signal.slot_count() + heartbeat_signal.slot_count();
+        std::unique_lock lock(_input_mutex);
```
weyrick: since this method is const, it can use shared_lock instead of unique_lock, for concurrent read access. i think a few other methods could have that change made too.

src/InputStream.h
```diff
@@ -10,20 +10,66 @@
 
 namespace visor {
 
+enum class Action {
+    AddPolicy,
+    RemovePolicy
+};
+
+class InputCallback : public Configurable
```
weyrick: can we please make this InputEventProxy instead of InputCallback

src/InputStream.h
```diff
@@ -51,20 +101,36 @@ class InputStream : public AbstractRunnableModule
...
-        return policy_signal.slot_count() + heartbeat_signal.slot_count();
+        std::unique_lock lock(_input_mutex);
+        size_t count = 0;
+        for (auto const &callback : _callbacks) {
+            count = callback->consumer_count();
+        }
+        return count;
     }
 
+    InputCallback *add_callback(const Configurable &filter)
```
weyrick: add_event_proxy

src/InputStream.h
```diff
@@ -51,20 +101,36 @@ class InputStream : public AbstractRunnableModule
...
+        for (auto const &callback : _callbacks) {
+            if (callback->hash() == hash) {
+                return callback.get();
+            }
+        }
+        _callbacks.push_back(create_callback(filter));
+        return _callbacks.back().get();
+    }
+
+    virtual std::unique_ptr<InputCallback> create_callback(const Configurable &filter) = 0;
```
weyrick: create_event_proxy

src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -229,30 +229,44 @@ void PcapInputStream::stop()
 
 void PcapInputStream::tcp_message_ready(int8_t side, const pcpp::TcpStreamData &tcpData)
 {
-    tcp_message_ready_signal(side, tcpData);
+    std::unique_lock lock(_input_mutex);
```
weyrick: ouch, it sucks to take a mutex lock hit here per event. we can make this shared_lock instead of unique lock though, since we are not mutating _callbacks. i think we'll need to benchmark this.

src/inputs/pcap/PcapInputStream.cpp
weyrick: why do we need to lock and unlock, is this recursive?

src/inputs/flow/FlowInputStream.cpp
weyrick: right. well, you can make all of the dynamic_casts static_casts i think, at least. should be faster and we know it will always be this class.

weyrick (changes requested):

src/inputs/pcap/PcapInputStream.cpp
leoparente: shared for those calls did improve the calls. did not see performance loss between current implementation and new implementation

src/InputStream.h
leoparente: done

src/inputs/pcap/PcapInputStream.cpp
leoparente: fixed that

src/InputStream.h
```diff
@@ -10,20 +10,66 @@
...
 class InputStream : public AbstractRunnableModule
 {
-    mutable std::shared_mutex _input_mutex;
+
     std::vector<const Policy *> _policies;
 
 protected:
     static constexpr uint8_t HEARTBEAT_INTERVAL = 30; // in seconds
+    mutable std::shared_mutex _input_mutex;
+    std::vector<std::unique_ptr<InputEventProxy>> _events;
```
weyrick: please make this `_event_proxies` rather than `_events` and use `proxy` for the loop variable rather than `event`

weyrick (changes requested):

src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -229,30 +229,41 @@ void PcapInputStream::stop()
 
 void PcapInputStream::tcp_message_ready(int8_t side, const pcpp::TcpStreamData &tcpData)
 {
-    tcp_message_ready_signal(side, tcpData);
+    for (auto &proxy : _event_proxies) {
```
weyrick, line 233: missing the shared lock

src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -229,30 +229,41 @@ void PcapInputStream::stop()
...
+        dynamic_cast<PcapInputEventProxy *>(proxy.get())->tcp_message_ready_cb(side, tcpData);
+    }
     _lru_list.put(tcpData.getConnectionData().flowKey, tcpData.getConnectionData().endTime);
 }
 
 void PcapInputStream::tcp_connection_start(const pcpp::ConnectionData &connectionData)
 {
-    tcp_connection_start_signal(connectionData);
+    for (auto &proxy : _event_proxies) {
```
weyrick, line 242: missing the shared lock

src/inputs/pcap/PcapInputStream.cpp
```diff
@@ -229,30 +229,41 @@ void PcapInputStream::stop()
...
+        dynamic_cast<PcapInputEventProxy *>(proxy.get())->tcp_connection_start_cb(connectionData);
+    }
     _lru_list.put(connectionData.flowKey, connectionData.startTime);
 }
 
 void PcapInputStream::tcp_connection_end(const pcpp::ConnectionData &connectionData, pcpp::TcpReassembly::ConnectionEndReason reason)
 {
-    tcp_connection_end_signal(connectionData, reason);
+    for (auto &proxy : _event_proxies) {
```
weyrick, line 251: missing the shared lock

src/Policies.cpp
```diff
@@ -146,6 +143,14 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
                 throw PolicyException(fmt::format("unable to instantiate tap '{}': {}", tap_name, e.what()));
             }
         }
+
+        InputEventProxy *input_event_proxy = nullptr;
+        try {
+            input_event_proxy = input_ptr->add_event_proxy(tap_filter);
+        } catch (ConfigException &e) {
+            throw PolicyException(fmt::format("invalid input filter config: {}", e.what()));
```
weyrick: should this error message be broader? "unable to create event proxy"? not sure.

weyrick (changes requested):

src/inputs/pcap/PcapInputStream.cpp
leoparente: tcp_message_ready and the others are only called by `auto result = _tcp_reassembly.reassemblePacket(packet);`(line 402)
of `process_raw_packet` which is already in a lock context (line 392). i did not what to add this overhead lock validation there.

leoparente: done

src/Policies.cpp
leoparente: done, please check

weyrick (approved):

PR #290, opened 2022-05-31
Add struct for network handler and count SYN packets

### LINKED ISSUE #53: NET module: counter for SYN packets

leoparente:


src/handlers/net/NetStreamHandler.h
```diff
@@ -58,6 +82,7 @@ class NetworkMetricsBucket final : public visor::AbstractMetricsBucket
             , OtherL4("packets", {"other_l4"}, "Count of packets which are not UDP or TCP")
             , IPv4("packets", {"ipv4"}, "Count of IPv4 packets")
             , IPv6("packets", {"ipv6"}, "Count of IPv6 packets")
+            , SYN("packets", {"syn"}, "Count of SYN packets")
```
weyrick: since SYN is TCP only, I recommend we call it TCP_SYN, and put it under a new metric heirarchy that's protocol specific: `{"protocol", "tcp", "syn"}` 

then we would have `{"packets": {"tcp": 100}, {"protocol": {"tcp": {"syn": 20}}}` in JSON output and

```
packets_tcp{module="default-net",policy="default"} 100
packets_protocol_tcp_syn{module="default-net",policy="default"} 20
```

in prometheus output... presumably we will expand the protocol specific section in the future.

thoughts @rboucher-ns1 ?

weyrick (changes requested):

src/handlers/net/NetStreamHandler.h
rboucher-me: I agree. This structure organizes the data in a way that is logical, and that will aid in the consumption of that data downstream (even programmatically).

My only concern is whether the TCP packets metric (for example) should _also_ exist under "protocol: tcp". For backwards compatibility we don't want to take it out from the "packets" level, but not duplicating it makes the "protocol: tcp" be a subset of all the actual available TCP metrics. 🤔

weyrick: Right so should we add packets.protocol.tcp.total which is a copy of packets.tcp  ... it's a good question. I think before doing this we need better metric documentation because the question will immediately arise when people look at the data "what's the difference" ... so let's leave that out for now, and add it later when we document metrics better. We'll definitely need to get good at evolving these schemas.

weyrick (approved):

PR #293, opened 2022-06-13
Multiply Sflow sampling rate by default

leoparente:


src/handlers/flow/FlowStreamHandler.h
```diff
@@ -184,6 +184,8 @@ class FlowStreamHandler final : public visor::StreamMetricsHandler<FlowMetricsMa
     std::vector<Ipv4Subnet> _IPv4_host_list;
     std::vector<Ipv6Subnet> _IPv6_host_list;
 
+    bool _sflow_sampling_rate;
```
weyrick: i think this probably applies to other flow types as well? so we should not make it sflow specific. and i propose we call it "sample_rate_scaling" @rboucher-ns1 ?

also, wouldn't other metrics besides payload size need this?

finally, i want to expose the sample rate as its own metric as well.

weyrick (changes requested):

src/handlers/flow/FlowStreamHandler.h
leoparente: we do have the _rate metrics which is basically this. It is because sflow we have the sampling rate and for netflow we already have the summary for packets and bytes exchanged between src and dst.

leoparente: but i do agree to let more generic name

rboucher-me: I agree and prefer the more generic "sample_rate_scaling" for the configuration setting.

I also agree that in would be great if we could expose the sample rate as a metric, but I'm not sure how to achieve that when we aggregate data at the edge from multiple flow sources that might be at different sampling rates.

leoparente: , _rate(FLOW_SCHEMA, {"rates", "pps"}, "Rate of combined flow packets per second")
        , _throughput("payload", {"rates", "bps"}, "Rate of combined flow bytes per second")
        
so for sflow it will be 1 (packet) * sampling rate (the number of packets that could be sampled)

currently this rate increses only 1 at time because we are considering only 1 packet for sflow. It is because sflow we have the sampling rate and for netflow we already have the summary for packets and bytes exchanged between src and dst.

weyrick: I'm lost on this last comment. @leoparente was that in response to @rboucher-ns1?

Exposing the sample rate metric isn't a problem in terms of time series - each agent will have its own label so be separate. I agree that _using_ that data from a TSDB query that needs to aggregate across all of those to be useful could be hard/not useful.

leoparente: What I meant is that we don't need a separated metric do get sampling rate because we already have it. But before we were considering sampling rate of 1 packet per flow for sflow data

weyrick: ok. I don't quite follow but let's go ahead and land this.

weyrick (approved):

src/handlers/flow/FlowStreamHandler.h
leoparente: ok, lets do it

rboucher-me: @weyrick My point was that the _same_ agent might be receiving sFlow streams from multiple sources, but with different sampling rates.

PR #295, opened 2022-06-15
O1 implementation for handlers signal

leoparente:


leoparente:
### **Premises**
1. **matcher** is the one that will look all or part of the traffic and call InputEventProxy `send_metadata` 
2. **matcher** needs to implement call logic and validate its _key metadata_
3. **match** is the one that will be called _**only if**_ its key value pair matches (O(1))
4. **match** handler will only be created if there is a matcher registered with same key
5. **match** and **matcher** must have the same schema (e.g. "dns")
6. **match** is code agnostic. It allows to connect **_custom_** handlers matchers with **_standard_** handlers match as long as they have the same schema and key metadata.

leoparente:
diagram drawio


Based on image above we have the call hierarchy for **DNS UDP** traffic:

1. **_All Traffic Policy DNS Handler_** will be the **matcher**, so it register itself as any other module to PcapInputEventProxy `udp_signal`. 
2. _**Google Policy and Amazon Policy DNS Handler**_  are **matches**, so instead of registering to `udp_signal`, they perform `PcapInputEventProxy::register_udp_predicate_signal` to register a callback to be called when the match occurs.
3. When a packet arrives and it is UDP packet,  `PcapInputEventProxy::process_udp_packet_cb` is called and it calls `udp_signal` which will trigger  _**All Traffic Policy DNS Handler**_  call.
4. As  _**All Traffic Policy Policy DNS Handler**_  is a **matcher**, it will call  `PcapInputEventProxy::send_metadata` that will enable the flag `_any_match`
5. After `udp_signal` finishes calling all registered signals,  `PcapInputEventProxy::process_udp_packet_cb`  method will check if there is any match and **_if so_**, it will call the _matches registered callbacks_ (O(1))

```
-process_udp_packet_cb
---udp_signal starts
-----send_metadata
-------udp_signal ends
---------any_match?
------------predicate_udp_signals find
---------------specific signal call
```

src/inputs/pcap/PcapInputStream.h
```diff
@@ -158,6 +176,33 @@ class PcapInputEventProxy: public visor::InputEventProxy
...
+        if (const auto it = _registered_matchers.find(key); it == _registered_matchers.end()) {
+            throw PcapException(fmt::format("there is no matcher registered with key [{}]", key));
+        } else if (it->second != schema) {
+            throw PcapException(fmt::format("matcher schema [{}] does not match predicate schema [{}], they must be the same.", it->second, schema));
+        }
+        auto hash = key + value;
+        predicate_udp_signals[hash].push_back(callback);
+    }
+
+    void send_metadata(const std::string &key, const std::string &value)
```
weyrick: when is this called? your notes say in the udp_signal but i don't see it. why do we need "metadata" and "any_match"?

weyrick:
is this still dependent on the order of the policies? i also cannot wrap my head around the names of the concepts here, i will try to think of different names that might be more intuitive.

src/inputs/pcap/PcapInputStream.h
leoparente: I've not added that call yet. but it would be in the DnsStreamHandler.cpp. Just added the overall idea.

Hmm I think we could use metadata.empty() and metada.clear() to remove any_match. I think it will not be a big change in the performance

leoparente:
> is this still dependent on the order of the policies? i also cannot wrap my head around the names of the concepts here, i will try to think of different names that might be more intuitive.

It still. @rboucher-ns1 would propose something at Handler level and not policy level. So we should discuss that

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -333,6 +345,45 @@ void DnsStreamHandler::info_json(json &j) const
...
     j[schema_key()]["xact"]["open"] = _metrics->num_open_transactions();
+    j[schema_key()]["predicate"]["enabled"] = _using_predicate_signals;
+}
+
+inline void DnsStreamHandler::_register_predicate_filter(std::string f_key, std::string f_value)
+{
+    if (!_udp_qname_predicate_signal && !_udp_qname_predicate_signal) {
+        // all DnsStreamHandler race to install this predicate, which is only installed once and called once per udp event
+        // it's job is to return the predicate "jump key" to call matching signals
+        _udp_qname_predicate = [this](pcpp::Packet &payload, PacketDirection dir, pcpp::ProtocolType l3, uint32_t flowkey, timespec stamp) -> std::string {
```
leoparente: This became specific for qname. The way I thought about `_register_predicate_filter` would work for different type of filters (Qname2, Rcode, etc), for each filter type we are going to have one of this Lambda;;

src/inputs/pcap/PcapInputStream.h
```diff
@@ -123,9 +146,45 @@ class PcapInputEventProxy: public visor::InputEventProxy
...
+        auto predicate_jump_key = full_predicate_key + conditional_value;
+        if (_udp_predicates.find(full_predicate_key) == _udp_predicates.end()) {
+            _udp_predicates[full_predicate_key] = predicate;
+        }
+        // now install the given conditional signal based on the jump key
+        // record the connection so we can remove it later when the handler disconnects
+        _udp_predicate_connections[handler_id] = _udp_predicate_signals[predicate_jump_key].connect(callback);
+    }
+
+    void unregister_udp_predicate_signal(const std::string &handler_id)
```
leoparente, line 204: My PR problem remains here:
If a policy that contains the caller `_udp_predicates` is removed, the others predicates will stop being called

Other point: we are not cleaning the `_udp_predicate_signals` map.

src/handlers/dns/DnsStreamHandler.cpp
weyrick: yes, though i think this is closer to next gen embedded stream processing framework, where everything is more granular - including small filtering lambdas. i think we can share almost all code between the filters though, with common methods.

src/inputs/pcap/PcapInputStream.h
weyrick: hm i think you're right but only because i kept your code of the predicate being a member variable. i think in your code it had to be (?), but in this strategy it does not - it can be static because it's stateless (with the exception of the dns layer cache code)

we clean it by removing the connection, only other thing to do is remove the entry if there are 0 connections?

src/handlers/dns/DnsStreamHandler.cpp
leoparente: cool, great

src/inputs/pcap/PcapInputStream.h
```diff
@@ -123,9 +146,45 @@ class PcapInputEventProxy: public visor::InputEventProxy
...
+        // first trigger generic udp signal
         udp_signal(payload, dir, l3, flowkey, stamp);
+
+        // if we have udp predicate signals, run each predicate and conditionally trigger signals that match
+        std::shared_lock lock(_pcap_proxy_mutex);
+        if (_udp_predicates.size()) {
+            for (const auto &[key, predicate] : _udp_predicates) {
+                auto predicate_jump_key = predicate(payload, dir, l3, flowkey, stamp);
+                if (_udp_predicate_signals.find(predicate_jump_key) != _udp_predicate_signals.end()) {
+                    auto i = _udp_predicate_signals[predicate_jump_key].slot_count();
```
leoparente: this `i` variable is not being used

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -376,6 +378,9 @@ class DnsStreamHandler final : public visor::StreamMetricsHandler<DnsMetricsMana
     mutable sigslot::signal<timespec> end_tstamp_signal;
     mutable sigslot::signal<const timespec> heartbeat_signal;
     mutable sigslot::signal<pcpp::Packet &, PacketDirection, pcpp::ProtocolType, uint32_t, timespec> udp_signal;
+
+    PcapInputEventProxy::UdpPredicate _udp_qname_predicate;
+    PcapInputEventProxy::UdpSignalCB _udp_qname_predicate_signal;
```
leoparente: this reference is not required anymore, right? we can only pass the lambda callback direct to the event proxy

src/inputs/pcap/PcapInputStream.h
leoparente: hmm I think we could change it to static, we just need to ensure that it exists for the filter key right? like "by_qname2"

src/handlers/dns/DnsStreamHandler.h
weyrick: yes i think that's right

src/inputs/pcap/PcapInputStream.h
weyrick: actually, on second thought - `_udp_predicates` is never removed from, even if policies are removed so i think this is ok. it's just the signal connections which are cleaned up, not the predicate.

src/inputs/pcap/PcapInputStream.h
weyrick: yes, was just for debug, can remove.

leoparente:
@weyrick  Just noticed that with that approach we loose the feature to track `filtered out counter` because the function that track it is never called

rboucher-me:
> @weyrick Just noticed that with that approach we loose the feature to track `filtered out counter` because the function that track it is never called

That's a small (and acceptable) price to pay

weyrick:
> > @weyrick Just noticed that with that approach we loose the feature to track `filtered out counter` because the function that track it is never called
> 
> That's a small (and acceptable) price to pay

Cool ok let's accept losing it for now then. I think there may be options in the future to add it back (based on failing predicates) but adds complication so let's avoid if we can.

src/inputs/pcap/PcapInputStream.h
```diff
@@ -123,9 +147,44 @@ class PcapInputEventProxy: public visor::InputEventProxy
         packet_signal(payload, dir, l3, l4, stamp);
     }
 
+    // dns, <handlerid> by_qname2, google.com, <func that returns current qname>, <func that processes udp>
```
weyrick: remove, this is debug really

weyrick (approved):
woot!

src/inputs/pcap/PcapInputStream.h
leoparente: done

PR #301, opened 2022-06-17
Add TopN based on Failed IPs on GeoLoc and ASN lookup

### LINKED ISSUE #302: NET module: Add handler filter for IPs that fail GeoLoc lookup

### LINKED ISSUE #303: NET module: Add handler filter for IPs that fail ASN lookup

leoparente:


src/GeoDB.cpp
```diff
@@ -42,7 +42,7 @@ MaxmindDB::~MaxmindDB()
     }
 }
 
-std::string MaxmindDB::getGeoLocString(const struct sockaddr *sa) const
+std::string MaxmindDB::getGeoLocString(const struct sockaddr *sa, bool &error) const
```
rboucher-me: instead of calling it `error`, should we call it `notfound`? it's not an actual error but rather a status/condition

src/handlers/flow/FlowStreamHandler.cpp
```diff
@@ -367,6 +367,8 @@ void FlowMetricsBucket::specialized_merge(const AbstractMetricsBucket &o)
     if (group_enabled(group::FlowMetrics::TopGeo)) {
         _topGeoLoc.merge(other._topGeoLoc);
         _topASN.merge(other._topASN);
+        _topGeoLocError.merge(other._topGeoLocError);
```
rboucher-me: similar to previous, should we consider calling it `_topGeoLocNotFound`?

src/handlers/flow/FlowStreamHandler.cpp
```diff
@@ -367,6 +367,8 @@ void FlowMetricsBucket::specialized_merge(const AbstractMetricsBucket &o)
...
+        _topASNError.merge(other._topASNError);
```
rboucher-me: `_topASNNotFound`?

src/GeoDB.cpp
```diff
@@ -127,7 +129,7 @@ std::string MaxmindDB::_getGeoLocString(MMDB_lookup_result_s *lookup) const
     return geoString;
 }
 
-std::string MaxmindDB::getASNString(const struct sockaddr *sa) const
+std::string MaxmindDB::getASNString(const struct sockaddr *sa, bool &error) const
```
rboucher-me: similar to previous, instead of calling it error should we call it `notfound`? it's not an actual error but rather a status/condition

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -133,6 +135,8 @@ class FlowMetricsBucket final : public visor::AbstractMetricsBucket
         , _dstPortCard(FLOW_SCHEMA, {"cardinality", "dst_ports_out"}, "Destination ports cardinality")
         , _topGeoLoc(FLOW_SCHEMA, "geo_loc", {"top_geoLoc"}, "Top GeoIP locations")
         , _topASN(FLOW_SCHEMA, "asn", {"top_ASN"}, "Top ASNs by IP")
+        , _topGeoLocError(FLOW_SCHEMA, "geo_loc", {"top_geoLoc_error"}, "Top IP addresses that failed GeoLoc lookup")
```
rboucher-me: "Top IP addresses that were not found during Geo Location lookup"

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -133,6 +135,8 @@ class FlowMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _topASNError(FLOW_SCHEMA, "asn", {"top_ASN_error"}, "Top IP addresses that failed ASN lookup")
```
rboucher-me: "Top IP addresses that were not found during ASN lookup"

src/handlers/net/NetStreamHandler.h
```diff
@@ -103,6 +105,8 @@ class NetworkMetricsBucket final : public visor::AbstractMetricsBucket
         , _dstIPCard("packets", {"cardinality", "dst_ips_out"}, "Destination IP cardinality")
         , _topGeoLoc("packets", "geo_loc", {"top_geoLoc"}, "Top GeoIP locations")
         , _topASN("packets", "asn", {"top_ASN"}, "Top ASNs by IP")
+        , _topGeoLocError("packets", "geo_loc", {"top_geoLoc_error"}, "Top IP addresses that failed GeoLoc lookup")
```
rboucher-me: "Top IP addresses that were not found during Geo Location lookup"

src/handlers/net/NetStreamHandler.h
```diff
@@ -103,6 +105,8 @@ class NetworkMetricsBucket final : public visor::AbstractMetricsBucket
...
+        , _topASNError("packets", "asn", {"top_ASN_error"}, "Top IP addresses that failed ASN lookup")
```
rboucher-me: "Top IP addresses that were not found during ASN lookup"

rboucher-me (commented):
Looks really good and thanks for taking this on so quickly (much appreciated!).

My main comment/concern is around using the term `error` in this context. I would rather we give the variables names that reflect the status of the lookup (like `notfound` or `lookupfailure`). My preference is for the shorter `notfound`.

weyrick:
hey let's chat about this first, i think we should do this differently

weyrick (changes requested):
agreed on calling it notfound. but i do not think this should be a new top list at all - this instead be a filter on notfound. then we get all metrics (including top ip and asn) just because we can filter on all notfounds.

rboucher-me:
@weyrick The reason I suggested a metric instead of a filter was so that it would be broadly available (doesn't require a specific policy) for NS1 customers of DNS Insights. I think these metrics could have very specific value-add for NS1 customers using Filter Chains based on ASN and/or GeoLoc

weyrick:
discussed in slack here https://ns1labs.slack.com/archives/C01UNG6UK37/p1655731526294639

cmd/pktvisord/main.cpp
```diff
@@ -66,6 +66,7 @@ static const char USAGE[] =
     Geo Options:
       --geo-city FILE             GeoLite2 City database to use for IP to Geo mapping
       --geo-asn FILE              GeoLite2 ASN database to use for IP to ASN mapping
+      --geo-cache N               GeoLite2 cache size for databases (default: 10000)
```
weyrick: `GeoLite2 LRU cache size, 0 to disable. (default: 10000)`

and i'd prefer `--geo-cache-size`

src/GeoDB.h
```diff
@@ -17,7 +20,7 @@ class MaxmindDB
 public:
     ~MaxmindDB();
 
-    void enable(const std::string &database_filename);
+    void enable(const std::string &database_filename, int cache_size = 10000);
```
weyrick: let's move that number to a const somewhere

weyrick (changes requested):
looks pretty good. i don't see anything that tests the filter count though? it seems to me the filter is checking != "Unknown" when it should check == so just want to make sure this has a test. 

this new LRU cache will affect the all Geo lookups, not just during filter right? i'd like to measure processing speed before and after this... this type of cache was on the optimization list anyway so nice to see it but let's prove effectiveness.

leoparente:
I've started doing some profiling using real scenario pcap data and database.

The criteria to profile were:
**1)** Debug build to check every method call 
**2)** use pktvisor-reader with same configuration in all profiling tests
**3)** Start process and wait at least 1M packets to be processed 

The result for `develop branch` was the following:
geo_develop
We can notice that _Net Handler Call_ used **43.6%** while DNS Handler call consumed 28.3% of process CPU.

If we take a look on Net Handler Calls, MaxmindDB calls consumes **28.7%** (City + ASN)
geo_dev_expand

leoparente:
Performed same profiling for `this branch` with `no cache`:

Testing it with cache size = 0, which will disable cache. so the expected is a similar result of develop.
geo_without_cache
And the similar result occurred as expected: **43.5%**

Same for Net Handler calls. DB calls consumed **28.5%**
geo_no_cache_expand

leoparente:
Now, profiling over `this branch` with the `LRUCache enabled ` (10k sized cache)

cache_geo
Here we can see a reduction of CPU consumption on Net Handler Call. It is consuming almost the same as DNS Handler call.

Looking at DB calls, it reduced to **17.7%** (more then 10%)
cache_expand

rboucher-me (approved):

weyrick:
nice. can we get an idea of how many entries that took in the cache, what the hit rate is, how much memory it takes?

leoparente:
## MEMORY

Doing some memory testing ( PS: debug build)

###  **10M Packets**

**Develop**
memory_10M_dev

**Cache 10k GeoDB branch**
memory_10M_cache

**Result:** We can see that the cache branch consumes 1MB more memory, however it takes **11 less seconds** to proccess 10M packets.

### 50M Packets

**Develop**
memory_50M_dev

**Cache 10k GeoDB branch**
memory_50M_cache

**Result:** We can see that the cache branch still consumes 1MB more then develop,  however it takes **almost 1 minute** less then develop to process 50M packets

weyrick (approved):
awesomeness

PR #304, opened 2022-06-20
Add global handler configuration at agent level

### LINKED ISSUE #207: Implement global handler configuration
As a pktvisor user with multiple dynamic policies, I want to be able to pre-configure some handler configuration per agent and not per policy.

For example, if a custom handler requires access to a file on disk at the agent (i.e. to lookup information per evend), I do not want to configure this directory per policy but rather once in the pktvisor agent configuration YAML, and have it apply to all instances of that handler on that agent. This configuration would be a default value which could be overridden per policy if it existed.

leoparente:


src/tests/test_policies.cpp
```diff
@@ -22,6 +22,8 @@ version: "1.0"
       input_type: mock
       config:
         iface: eth0
+  global_handler_config:
```
weyrick, line 25: We need to update the policy RFC with this specification. That RFC already specifies a section for global handler config under "visor.policies.handler.config", which i think we could adjust a bit to add "global" and handler type specific config like

```
      handlers:
        # default configuration for the stream handlers
        config:
           global:
             periods: 5
             max_deep_sample: 50
           type:
              dns:
                 only_rcode: 2
          
```

weyrick (changes requested):

src/tests/test_policies.cpp
leoparente: I don't get it. I thought that the idea is to **make it agent configurable at startup level** and not at Policy level.
Because at policy level, as it is global, a policy can apply that after another policy is applied and next policy will use this new configuration.

Also when the agent reboots, the applied policy order can change and it can cause different behaviors  every reboot

weyrick: You're right sorry, but it does need to be in rfc and it should still be broken out per handler.

weyrick: Oh i guess we don't have an rfc for config! We should fix that too :P

leoparente: > but it does need to be in rfc and it should still be broken out per handler.

could you explain better the broken out per handler?

weyrick: yes right now you have a section which will apply key/val config to _all_ handlers but i want to be able to specify global key/val config per handler. so instead of

```
visor:
  taps:
    anycast:
      input_type: mock
      config:
        iface: eth0
  global_handler_config:
    only_rcode: 2
```
it should be 

```
visor:
  taps:
    anycast:
      input_type: mock
      config:
        iface: eth0
  global_handler_config:
    dns:
      only_rcode: 2
```

leoparente: done. We should discuss better how to add a RFC for this and also for Config. It should be two separated RFCs? 
if so, what should be the name of them?

weyrick: no just a single RFC for config file schema. it should follow the pattern of the other RFCS - create an issue for it, then create the RFC file name with the data and issue number in it.

we can land this though and do that separately.

weyrick (approved):

src/tests/test_policies.cpp
leoparente: issue created: https://github.com/ns1labs/pktvisor/issues/307

I will land this one

PR #308, opened 2022-06-23
Feature/configurable topn count

### LINKED ISSUE #300: Top 10 metrics: Make "10" a configurable parameter
Ideally this would be configurable at the individual Top 10 metric, but an initial implementation could be limited to setting the value at the module level in the policies (i.e. per named handler)

leoparente:


leoparente:
@rboucher-ns1  i did it per policy at same level of deep sample and periods should it be at each handler level?
@weyrick what are your thoughts?

rboucher-me:
The _ideal_ scenario would be to set it at the individual metric level in the handler configuration (`dns_top_qnames3_items=20`). Second-best would be at the handler level (`top_items=20`).
That said, I think setting this at the policy level is sufficient. The purpose of this is to allow us to do some testing and tuning (particularly for flow metrics) to find the ideal compromise between data granularity and compute/storage requirements.

rboucher-me (approved):

src/Metrics.h
```diff
@@ -259,6 +259,16 @@ class TopN final : public Metric
         _fi.merge(other._fi);
     }
 
+    void update_top_count(const size_t top_count)
```
weyrick: the convention is to prepend `set_` for setters, and for accessors to have the same name as the variable without the prepended underscore, so these should be `set_top_count` and `top_count`

RFCs/2021-04-16-76-collection-policies.md
```diff
@@ -43,8 +43,9 @@ visor:
       handlers:
         # default configuration for the stream handlers
         config:
-          periods: 5
-          max_deep_sample: 50
+          periods: 2 #default is 5
+          max_deep_sample: 50 #default is 100
+          topn_size: 5 #default is 10
```
weyrick: let's stick with "count" instead of "size", as per the existing metrics member variable

src/AbstractMetricsManager.h
```diff
@@ -238,6 +239,7 @@ class AbstractMetricsManager
      */
     jsf32 _rng;
     uint32_t _deep_sample_rate{100};
+    size_t _topn_size{10};
```
weyrick: here as well, use _count

weyrick (changes requested):

RFCs/2021-04-16-76-collection-policies.md
leoparente: i've made `topn_count` here. Do you think that it should be `top_count` (without the 'n')

src/Metrics.h
leoparente: done

RFCs/2021-04-16-76-collection-policies.md
weyrick: topn_count is good

src/AbstractMetricsManager.h
```diff
@@ -214,6 +214,7 @@ class AbstractMetricsBucket
 
     virtual void to_json(json &j) const = 0;
     virtual void to_prometheus(std::stringstream &out, Metric::LabelMap add_labels = {}) const = 0;
+    virtual void update_top_metrics(size_t top_count) = 0;
```
weyrick: let's do `topn_count` across the board, so `update_topn_metrics(size_t topn_count)`

src/Metrics.h
```diff
@@ -259,6 +259,16 @@ class TopN final : public Metric
         _fi.merge(other._fi);
     }
 
+    void set_top_count(const size_t top_count)
```
weyrick: `set_topn_count`

src/Metrics.h
```diff
@@ -259,6 +259,16 @@ class TopN final : public Metric
...
+    {
+        _top_count = top_count;
+    }
+
+    size_t get_top_count() const
```
weyrick: just `size_t topn_count() const`, no "get_"

weyrick (changes requested):

src/Metrics.h
leoparente: done

src/Metrics.h
leoparente: done

weyrick (approved):

PR #310, opened 2022-06-27
Replace int with string on thread monitor stat collection

### LINKED ISSUE #306: ThreadMonitor failing on CentOS7
Should test it with docker image.

leoparente:


leoparente:
`thread_stat.ignore(' ');` was causing different behaviors, and stat has string and int values. So the best approach is to read everything as string and then convert the necessary values
  
  sample stats:
  ```
  98 (cat) R 18 198 18 34817 198 4194304 78 0 0 0 0 0 0 0 20 0 1 0 1299013 4505600 185 18446744073709551615 4194304 4238828 140735373488928 0 0 0 0 0 0 0 0 0 17 6 0 0 0 0 0 6339656 6341408 19660800 140735373490557 140735373490584 140735373490584 140735373492203 0
```

File: src/handlers/input_resources/ThreadMonitor.h [hunk 1]
```diff
@@ -46,17 +48,20 @@ class ThreadMonitor
...
-        std::vector<uint64_t> stats;
+        std::vector<std::string> stats;
         std::ifstream thread_stat("/proc/thread-self/stat");
-        thread_stat.ignore(' ');
-        while (thread_stat >> stat) {
-            stats.push_back(stat);
+        std::string stat_str;
+        while (thread_stat >> stat_str) {
+            stats.push_back(stat_str);
+            if (stats.size() > PROC_STAT_POS_STIME)
```
leoparente, line 56: proc stat has more than 40 items and we only use the 13 and 14, so I think it is less **costly** to check size 15 times than pushing back string 25 times without being used

weyrick (approved):

PR #316, opened 2022-06-30
Add NODATA count and top qname

### LINKED ISSUE #298: DNS module: Total DNS NODATA wire packets (return code NOERROR and no answer in response)
This issue is to add a new DNS metric that tracks NODATA requests (return error code NOERROR and no answer section in the response).

### LINKED ISSUE #299: DNS module: Top NODATA QNAMES (result code NOERROR and no answer in response)
Add to DNS handler Top NODATA QNAMES (result code NOERROR and no answer in response). This is related to issue pktvisor:298

leoparente:


src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -71,6 +71,10 @@ void DnsStreamHandler::start()
             throw ConfigException("only_rcode contained an invalid/unsupported rcode");
         }
     }
+    if (config_exists("answer_count")) {
```
weyrick, line 79: i think we need to also ensure it's an int, as we do on line 57, otherwise 75 will fail

weyrick (changes requested):

src/handlers/dns/DnsStreamHandler.cpp
leoparente: but i did the same as line 60 and 61. it does not ensure if it is an int before calling get

weyrick: hm isn't that a bug too though? what actually happens in case the type is wrong?

leoparente: policy apply will fail. I thought that is the expected behavior. because it the type is wrong we should throw right? Which is that we currently do

leoparente: we also throw when rcode is not in the known range for instance

weyrick: yes if we handle the error gracefully then that's correct - but also the error message needs to be understandable so they can know what to fix too. remember this error message will show up in the orb UI

leoparente: Cool. I will work on that!

leoparente: done

weyrick (approved):

Issue #312, opened 2022-06-29
Multiple Handler Support in pktvisor


rboucher-me:
pktvisor should support multiple handlers defined in a policy, including:
- ensuring multiple handlers of the same type (inside the same policy) don't conflict
- ensuring the handler name/label is included in the exposed metrics

leoparente:
Performed some testing on this task. On Pktvisor side everything seems to be **_working properly_**:

**Policy used:**
```
version: "1.0"

visor:
  config:
    admin_api: true
  taps:
    default:
      input_type: pcap
      config:
        iface: wlp0s20f3
  policies:
   default:
      kind: collection
      input:
        tap: default
        input_type: pcap
        config:
          host_spec: "198.51.44.0/23,2a00:edc0:6259::/48,2620:4d:4000::/48"
      handlers:
        window_config:
          num_periods: 5
          deep_sample_rate: 100
        modules:
          net:
            type: net
          net2:
            type: net
          dns1:
            type: dns
          dns2:
            type: dns
```

**JSON Result** (`/api/v1/policies/__all/metrics/bucket/1` -> route used by ORB)
```
{
   "default":{
      "default-dns1":{
         "dns":{
            "cardinality":{
               "qname":1
            },
            "period":{
               "length":60,
               "start_ts":1656682330
            },
            "rates":{
               "total":{
                  "p50":0,
                  "p90":1,
                  "p95":1,
                  "p99":2
               }
            },
            "top_nxdomain":[
               
            ],
            "top_qname2":[
               {
                  "estimate":3,
                  "name":"._tcp.local"
               }
            ],
            "top_qname3":[
               {
                  "estimate":3,
                  "name":"._googlecast._tcp.local"
               }
            ],
            "top_qtype":[
               {
                  "estimate":3,
                  "name":"PTR"
               }
            ],
            "top_rcode":[
               {
                  "estimate":15,
                  "name":"NOERROR"
               }
            ],
            "top_refused":[
               
            ],
            "top_srvfail":[
               
            ],
            "top_udp_ports":[
               {
                  "estimate":18,
                  "name":"5353"
               }
            ],
            "wire_packets":{
               "deep_samples":18,
               "filtered":0,
               "ipv4":18,
               "ipv6":0,
               "noerror":15,
               "nxdomain":0,
               "queries":3,
               "refused":0,
               "replies":15,
               "srvfail":0,
               "tcp":0,
               "total":18,
               "udp":18
            },
            "xact":{
[...]

PR #318, opened 2022-07-01
Handle error properly when handler module add fails

leoparente:


leoparente:
@weyrick testing https://github.com/ns1labs/pktvisor/issues/312 I notice that error handler was not being properly done when two handlers has the same name. It was a side effect of the reuse of input stream and resource usage policy implementation. This is a fix to properly handle it.

weyrick (approved):

PR #321, opened 2022-07-04
Add own implementation for public suffix list

### LINKED ISSUE #232: DNS stats should be based upon public suffix

leoparente:


rboucher-me (approved):
Did not validate the code itself, but the implementation looks good to me.

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -335,14 +334,18 @@ bool DnsStreamHandler::_filtering(DnsLayer &payload, [[maybe_unused]] PacketDire
...
+            if (ends_with(qname_ci, fqn)) {
                 _static_suffix_size = fqn.size();
                 goto will_not_filter;
             }
         }
         // checked the whole list and none of them matched: filter
         goto will_filter;
     }
+    // This is not actually a filter, but only should work if OnlyQNameSuffix is not enabled
+    else if (_f_enabled[Filters::PublicSuffixList] && payload.parseResources(true) && payload.getFirstQuery() != nullptr) {
```
weyrick: did you use _f_enabled for convenience? i would rather not conflate it with config options which are not filters. if it makes sense we can add _c_enabled and a similar list of config functionality

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -335,14 +334,18 @@ bool DnsStreamHandler::_filtering(DnsLayer &payload, [[maybe_unused]] PacketDire
...
+        _static_suffix_size = match_public_suffix(payload.getFirstQuery()->getNameLower());
```
weyrick: remind me - we cache this lower name right?

weyrick (changes requested):

src/handlers/dns/DnsStreamHandler.cpp
leoparente: yes, we do

src/handlers/dns/DnsStreamHandler.cpp
leoparente: yes. i did for convenience. I will change that

leoparente: done

weyrick (approved):

PR #323, opened 2022-07-06
Feature/qname response size

### LINKED ISSUE #313: Add Top QNAMES by Response Size, Query/Reply size ratio metrics
This ticket is to add a new Top metric for QNAMES by Response Size, using the size of the DNS packet as the weight in the frequent items sketch.

In addition, add a new histogram metric for tracking the ratio between DNS Queries and Replies, which we can track in the DNS transaction code.

The goal is to be able to monitor for amplification attacks.

leoparente:


src/handlers/dns/DnsStreamHandler.h
```diff
@@ -53,6 +53,7 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
 
     Quantile<uint64_t> _dnsXactFromTimeUs;
     Quantile<uint64_t> _dnsXactToTimeUs;
+    Quantile<uint64_t> _dnsXactRatio;
```
weyrick: let's make a double (or float?)

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -127,6 +130,7 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
         , _dns_topQname3("dns", "qname", {"top_qname3"}, "Top QNAMES, aggregated at a depth of three labels")
         , _dns_topNX("dns", "qname", {"top_nxdomain"}, "Top QNAMES with result code NXDOMAIN")
         , _dns_topREFUSED("dns", "qname", {"top_refused"}, "Top QNAMES with result code REFUSED")
+        , _dns_topSizedQnameResp("dns", "qname", {"top_qname_by_response"}, "Top QNAMES by response size")
```
weyrick: i think we need `top_qname_by_response_size` ... possibly `top_qname_by_resp_size` if we don't have another precedent?

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -119,6 +121,7 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
     DnsMetricsBucket()
         : _dnsXactFromTimeUs("dns", {"xact", "out", "quantiles_us"}, "Quantiles of transaction timing (query/reply pairs) when host is client, in microseconds")
         , _dnsXactToTimeUs("dns", {"xact", "in", "quantiles_us"}, "Quantiles of transaction timing (query/reply pairs) when host is server, in microseconds")
+        , _dnsXactRatio("dns", {"xact", "ratio", "quantiles_percent"}, "Quantiles of transaction ratio (query/reply pairs) in percentage")
```
weyrick: i guess just `quantiles` once we switch

weyrick (changes requested):

src/handlers/dns/DnsStreamHandler.h
leoparente: ok, what about `_dns_topSizedQnameResp`? is it a good name?

src/handlers/dns/DnsStreamHandler.h
leoparente: done

src/handlers/dns/DnsStreamHandler.h
leoparente: made it double. done

src/handlers/dns/DnsStreamHandler.h
weyrick: yeah it's ok, the internal name i don't care as much about since it's not a public interface, we can always refactor it later

leoparente: cool

leoparente:
@weyrick `top_qname_by_resp_size` should be **full dns qname** or **qname2**

I did for full qname, like NoDATA and SRVFAIL

weyrick (approved):

PR #324, opened 2022-07-07
Fix thread unsafe issue on GeoDB

leoparente:


leoparente:
Did some profiling with mutex and things did not change that much.
image

weyrick (approved):

weyrick (approved):

PR #329, opened 2022-07-11
buckets_queue_metrics

manrodrigues:
the idea here is to validate that the data that is in bucket 1 at minute 1 is in bucket 2 at minute two, etc.
we can define the best file to use.. let's evolve this test to tests of expected metrics;


etaques (approved):
LGTM

PR #330, opened 2022-07-12
Implement tap selector

### LINKED ISSUE #266: RFC: Tap selectors
Currently (RFC 76) a policy must specify exactly one Tap by name to create a source input stream from. This effectively binds the policy logic to a specific Tap entity. If the same policy logic should apply to a different Tap on the same agent, a new policy is required.

For use cases where multiple similar taps exist on a single agent (for example, several pcap taps with different BPF which shard the traffic capture streams, or on different physical interfaces on a switch/router), it would be convenient to specify _tap selectors_ instead of a single, explicit Tap. In this way a single policy could apply to multiple Taps on the agent.

A "tap selector", if specified instead of an explicit tap name, would be a set of key/value pairs which should match against key/value pairs listed in the Tap definition (RFC 75). The policy would then apply to each Tap which matched.

```
visor:
  taps:
    vhost1:
      input_type: pcap
      config:
        iface: eth0
        bpf: "host 192.168.0.1"
      tags:
        virtual: true
        vhost: 1
    vhost2:
      input_type: pcap
      config:
        iface: eth0
        bpf: "host 192.168.0.2"
      tags:
        virtual: true
        vhost: 2
  policies:
    sample_policy:
      kind: collection
      input:
        tap_selector:
          virtual:true
        input_type: pcap
      handlers:
...
```

leoparente:


leoparente:
Changed to draft: still validating manually it as well

leoparente:
Ready to review

RFCs/2021-04-16-76-collection-policies.md
```diff
@@ -84,8 +84,12 @@ visor:
       description: "base chaning NET to DNS policy"
       # input stream to create based on the given tap and optional filter config
       input:
-        # this must reference a tap name, or application of the policy will fail
-        tap: anycast
+        # this must reference valid tags existent on applied taps, or application of the policy will fail
+        tap_selector:
+          # It can be either "any" or "all"
```
weyrick, line 89: i think we probably want an optional `input_type` here to ensure a tap_selector only matches tags against a certain input source, if that makes sense for this policy. it may not always - for example, a dns policy that should work on both pcap and dnstap. but you may have flow policy that should never match against anything but flow input.

src/Taps.cpp
```diff
@@ -76,4 +83,48 @@ std::unique_ptr<InputStream> Tap::instantiate(const Configurable *config, const
     return module;
 }
 
+bool Tap::tags_validate_yaml(const YAML::Node &tag_yaml, bool all)
+{
+    bool any_match = false;
+    for (YAML::const_iterator it = tag_yaml.begin(); it != tag_yaml.end(); ++it) {
+        if (!it->second.IsScalar()) {
+            throw PolicyException(fmt::format("tag key '{}' must have scalar value", it->first));
```
weyrick: think we probably need a TapException and not PolicyException because tags are set at pktvisor config load time, not policy

src/Taps.h
```diff
@@ -32,6 +34,8 @@ class Tap : public AbstractModule
 
     std::unique_ptr<InputStream> instantiate(const Configurable *config, const Configurable *filter, std::string input_name);
 
+    bool tags_validate_yaml(const YAML::Node &tag_yaml, bool all);
```
weyrick: i think we are not validating here, but matching, right? so `tags_match_selector_yaml` maybe.

but i actually wonder if we should instead make this a method of Configurable. then we could use config_set_yaml, and we'd be able to reuse this matching functionality

src/Policies.cpp
```diff
@@ -96,17 +96,79 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
+                    auto policy_selector_name = policy_name + "_" + tmod->name();
+                    if (module_exists(policy_selector_name)) {
+                        throw PolicyException(fmt::format("policy with name '{}' already defined", policy_selector_name));
+                    }
+                    result.push_back(create_policy(it->second, policy_selector_name, tmod));
+                    match = true;
+                }
+            }
+            if (!match) {
+                throw PolicyException("no tap match found for specified 'input.tap_selector' tags");
```
weyrick: so there's a product question of whether it should be an error or not when a policy with tap_selectors has been sent to an agent but does not match any tap ... or indeed whether there should be a third state for "no match" (in addition to the states "applied" or "failed to apply"). @rboucher-ns1 any thoughts? we can discuss use cases

src/Policies.cpp
```diff
@@ -96,17 +96,79 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
+                all = false;
+                binary_op = "any";
+            }
+
+            bool match = false;
+            auto [tap_modules, hm_lock] = _registry->tap_manager()->module_get_all_locked();
+            for (auto &[name, mod] : tap_modules) {
+                auto tmod = dynamic_cast<Tap *>(mod.get());
+                if (tmod && tmod->tags_validate_yaml(tap_selector[binary_op], all)) {
+                    auto policy_selector_name = policy_name + "_" + tmod->name();
```
weyrick: I think we are going to have to find a way to make this policy name stay the same, but have it have an additional label for the tap it used. if we don't do this, it will be poor UX experience in the dashboards, with an ever expanding list of these policy-tap prefixes in the Policy dropdown

I believe some of our other use cases are also forcing us to think about the label issue, so hopefully there is a common solution

src/Policies.cpp
```diff
@@ -96,17 +96,79 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
+            if (!match) {
+                throw PolicyException("no tap match found for specified 'input.tap_selector' tags");
+            }
+        }
+    }
+
+    return result;
+}
+
+Policy *PolicyManager::create_policy(const YAML::Node &policy_yaml, std::string policy_name, Tap *tap)
```
weyrick: I like that you broke this out, but wonder if we can break it out even more - it's still very large, hard to understand method

weyrick (changes requested):

RFCs/2021-04-16-76-collection-policies.md
leoparente: we still validate the  `input type` for each match selector, `input type` still in RFC as well

leoparente: https://github.com/ns1labs/pktvisor/blob/8e41c48e716836d6bfe97bdc132f6bc38354e37b/src/Policies.cpp#L226

src/Taps.cpp
leoparente: agreed

RFCs/2021-04-16-76-collection-policies.md
weyrick: hm ok that precludes the pcap/dnstap use case i mentioned, but let's leave in place for now, we can always relax it later

leoparente: cool

src/Policies.cpp
leoparente: In any case, we should also think about JSON return, how labels will work for it

src/Taps.cpp
leoparente: done

src/Policies.cpp
leoparente: I broke out the handler validation as well

src/Taps.h
leoparente: renamed it

src/Policies.cpp
leoparente: @rboucher-ns1 any thoughts?

src/Policies.cpp
weyrick: @rboucher-ns1 let's discuss this, I think it's the only thing blocking landing this now.

weyrick: @leoparente richard and I discussed - we do think think this should be the same policy name but with a different top level label applied to all metrics. i think this means we need to add support for _always_ adding a "tap" label to all of our metric output, even when tap selectors aren't used

leoparente:
Tap name concatenated  to handler name to ensure uniqueness
```
{
    "default": {
        "default-default-dns1": {
            "dns": {
                "cardinality": {
                    "qname": 0
                },
                "period": {
                    "length": 1,
```

And it was added as a label for prometheus return value:
```
# HELP dns_rates_total Rate of all DNS wire packets (combined ingress and egress) per second
# TYPE dns_rates_total summary
dns_rates_total{module="default-default-dns1",policy="default",quantile="0.5",tap="default"} 0
dns_rates_total{module="default-default-dns1",policy="default",quantile="0.9",tap="default"} 0
dns_rates_total{module="default-default-dns1",policy="default",quantile="0.95",tap="default"} 1
dns_rates_total{module="default-default-dns1",policy="default",quantile="0.99",tap="default"} 1
```

src/Policies.cpp
leoparente: done

src/Policies.cpp
```diff
@@ -117,277 +192,285 @@ std::vector<Policy *> PolicyManager::load(const YAML::Node &policy_yaml)
...
-        bool handler_sequence = false;
-        if (!handler_node["modules"] || (!handler_node["modules"].IsMap() && !handler_node["modules"].IsSequence())) {
-            throw PolicyException("missing or invalid handler modules at key 'modules'");
-        } else if (handler_node["modules"].IsSequence()) {
-            handler_sequence = true;
+    } else {
+        window_config.config_set<uint64_t>("num_periods", _default_num_periods);
+        window_config.config_set<uint64_t>("deep_sample_rate", _default_deep_sample_rate);
+    }
+    window_config.config_set<std::string>("_internal_tap_name", tap_name);
```
leoparente, line 288: @weyrick I've added this internal config to pass the tap name to abstractManager so I did not need to add new `tap_name` parameter to Handlers constructor. What do you think?

leoparente:
Added error 422 when there is no match for tap selector:
image

src/Policies.cpp
leoparente: done

src/Policies.h
```diff
@@ -50,12 +48,22 @@ class Policy : public AbstractRunnableModule
         return "policy";
     }
 
+    void set_modules_sequence(bool sequence)
+    {
+        _modules_sequence = sequence;
+    }
+
+    void set_tap(Tap *tap)
```
weyrick: since it's a vector let's use `add_tap`

src/Policies.h
```diff
@@ -50,12 +48,22 @@ class Policy : public AbstractRunnableModule
...
+    {
+        _tap.push_back(tap);
+    }
+
     void set_input_stream(InputStream *input_stream)
     {
-        _input_stream = input_stream;
+        _input_stream.push_back(input_stream);
```
weyrick: similarly, should be `add_input_stream` now

src/Policies.h
```diff
@@ -50,12 +48,22 @@ class Policy : public AbstractRunnableModule
...
     }
 
-    const InputStream *input_stream() const
+    const std::vector<InputStream *> &input_stream()
```
weyrick: can't this method stay `const`?

src/Policies.h
```diff
@@ -31,17 +31,15 @@ class Policy : public AbstractRunnableModule
 {
     static constexpr size_t HANDLERS_SEQUENCE_SIZE = 1;
 
-    Tap *_tap;
+    std::vector<Tap *> _tap;
```
weyrick: i _think_ precedent is that this and _input_stream should be plural now, so `_taps`? let's see what we do elsewhere if you haven't

weyrick (changes requested):
so i think it's almost there - but it's also an extremely hard PR to review since lots of code is moved around and it's tricky logic. i think this means we _really_ need good unit tests because this is a critical section of code. i'd also love to find a way to simplify this policy loading code. ideally that means breaking it up into separate methods performing sections of the transaction which are no more than 20-30 lines long, say. but i don't know how realistic that is.

src/Policies.h
leoparente: yes, I've changed it

src/Policies.h
leoparente: done

src/Policies.h
leoparente: done

src/Policies.h
leoparente: done

leoparente:
> so i think it's almost there - but it's also an extremely hard PR to review since lots of code is moved around and it's tricky logic. i think this means we _really_ need good unit tests because this is a critical section of code. i'd also love to find a way to simplify this policy loading code. ideally that means breaking it up into separate methods performing sections of the transaction which are no more than 20-30 lines long, say. but i don't know how realistic that is.

Agreed. That's why I added new unit tests to validate it, did this based on Code Coverage to ensure to test new code.

I think this PR is a improvement on that policy load which add new two methods to break it a bit. I also agree that we can improve more that... But I think we could do that in another PR

weyrick (approved):

PR #347, opened 2022-07-27
Improve flow handler metrics

### LINKED ISSUE #344: Flow Handler - Create new metrics structure
This ticket is to add Flow handler support in pktvisor for the new set of metrics identified. Metrics should now be calculated based on the device ("device" label) the flows were received from, the  interface ("ifIndex" label, extracted from within the flow contents) and the interface direction ("direction" label, with possible values "in" and "out").

**Aggregate flow stats by interface**
flow_(packets|bytes)
Labels: “device”, “interface”, “direction”

**Top Talker stats by interface**
flow_top_(src|dst)_ip_(packets|bytes)
flow_cardinality_(src|dst)_ip

**Protocol stats by interface**
flow_top_(src|dst)_port_(packets|bytes)
flow_cardinality_(src|dst)_port

**Application stats by interface**
flow_top_(src|dst)_ip:port_(packets|bytes)
flow_cardinality_(src|dst)_ip:port

**Conversation stats**
flow_top_conversations_(packets|bytes)
Flow_cardinality_conversations

**Filters**
The handler should support the following filters:
- by IPs or Subnets (CIDR)
- by port number (or protocol name)
- by application (IP:port, Subnet:port-range pairs)
- Geo Location not found
- ASN not found

The handler should ideally provide metrics on itself:
- How many flows are processed
- How much flow volume (total payload size)
- How much metric volume are we generating

leoparente:


weyrick (approved):
did a fairly cursory review and looks pretty good to me. i'm ok to land this quicker than some of the other large PRs since it doesn't affect current production agents - let's start testing

PR #354, opened 2022-08-02
DnsHandler: Implement ECS counter and Geo and ASN Not Found filters

### LINKED ISSUE #338: DNS Handler - Add a NotFound filter on GeoLoc and ASN for ECS Addresses
This request is similar to the NET Handler notFound filter for GeoLoc and ASN databases, but instead for notFound ECS IPs (instead of source IPs).

### LINKED ISSUE #339: DNS Handler - Add metric for the number of inbound DNS queries with ECS set
This metric should provide a count of the number of DNS queries have the EDNS Client Subnet (ECS) field set. This will allow customers to know how much of their inbound DNS traffic can be steered based on this field.

leoparente:


src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -487,6 +519,7 @@ void DnsMetricsBucket::to_json(json &j) const
     _dns_topUDPPort.to_json(j, [](const uint16_t &val) { return std::to_string(val); });
 
     if (group_enabled(group::DnsMetrics::TopEcs)) {
+        group_enabled(group::DnsMetrics::Counters) ? _counters.queryECS.to_json(j) : void();
```
leoparente, line 578: I made this `queryECS` counter metric depend on both `TopEcs` and `Counters` because it needs a deep_sample (processing ECS packets) to be counted..., thoughts @weyrick ?

weyrick: yeah makes sense

src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -349,17 +355,42 @@ inline bool DnsStreamHandler::_filtering(DnsLayer &payload, [[maybe_unused]] Pac
...
+                    _static_suffix_size = fqn.size();
+                    return true;
+                }
+                return false;
+            })) {
+            // checked the whole list and none of them matched: filter
+            goto will_filter;
+        }
+    }
+    if (_f_enabled[Filters::GeoLocNotFound]) {
```
weyrick, line 425: these assume we're doing geo/asn on ECS only ... but we may want this filter to fallback to query source IP if ECS isn't available. @rboucher-ns1 ?

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -112,6 +114,7 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
             , NOERROR("dns", {"wire_packets", "noerror"}, "Total DNS wire packets flagged as reply with return code NOERROR (ingress and egress)")
             , NODATA("dns", {"wire_packets", "nodata"}, "Total DNS wire packets flagged as reply with return code NOERROR and no answer section data (ingress and egress)")
             , filtered("dns", {"wire_packets", "filtered"}, "Total DNS wire packets seen that did not match the configured filter(s) (if any)")
+            , queryECS("dns", {"wire_packets", "query_ecs"}, "Total of queries that have EDNS Client Subnet (ECS) field set")
```
weyrick: "Total queries that ..."

weyrick (changes requested):

src/handlers/dns/DnsStreamHandler.h
leoparente: done

src/handlers/dns/DnsStreamHandler.cpp
leoparente: cool!

src/handlers/dns/DnsStreamHandler.cpp
leoparente: That would already be achieved on a NetHandler right?

rboucher-me: @leoparente That was my thought as well. It's not ideal (from a user's perspective) that the metrics and filters associated with source IPs and ECS live in two different handlers, but I don't see an elegant way around that.

leoparente: @rboucher-ns1 but should we land it and then have a ticket to find a way to do that in an elegant way?

rboucher-me: I think these are two distinct filters, on two different handlers. I prefer them to stay separate so that a user can configure a policy to achieve his specific intent (i.e. whether to filter on notFound on source IP or ECS, or both).

What might be interesting in the future is chaining from NET handler to DNS handler, so that you could get DNS metrics on _only_ the traffic that has failed the GeoLoc or ASN lookup.

src/handlers/dns/DnsStreamHandler.h
```diff
@@ -112,6 +114,7 @@ class DnsMetricsBucket final : public visor::AbstractMetricsBucket
             , NOERROR("dns", {"wire_packets", "noerror"}, "Total DNS wire packets flagged as reply with return code NOERROR (ingress and egress)")
             , NODATA("dns", {"wire_packets", "nodata"}, "Total DNS wire packets flagged as reply with return code NOERROR and no answer section data (ingress and egress)")
             , filtered("dns", {"wire_packets", "filtered"}, "Total DNS wire packets seen that did not match the configured filter(s) (if any)")
+            , queryECS("dns", {"wire_packets", "query_ecs"}, "Total queries that have EDNS Client Subnet (ECS) field set")
```
rboucher-me, line 117: @leoparente Is it the "Total queries that..." or the "Total DNS wire packets that..."? Will it count both DNS queries and response packets with ECS set?

leoparente: only queries @rboucher-ns1. We are not tracking replies with ecs yet

rboucher-me (approved):

PR #355, opened 2022-08-03
Implement DNSSEC Filter

### LINKED ISSUE #314: Add DNS Handler Filter for DNSSEC Signed Responses
This ticket is to add support for a filter in the DNS handler to restrict analysis to only DNSSEC-Signed responses. Valid DNSSEC Signed DNS query responses include an RRSIG record in the response.

leoparente:


src/handlers/dns/DnsStreamHandler.cpp
```diff
@@ -76,6 +76,9 @@ void DnsStreamHandler::start()
             throw ConfigException("DnsStreamHandler: only_rcode filter contained an invalid/unsupported rcode");
         }
     }
+    if (config_exists("only_dns_sec_response")) {
```
weyrick: needs to be "only_dnssec_response" with no _ between dns and sec

weyrick (changes requested):

src/handlers/dns/DnsStreamHandler.cpp
leoparente: done

weyrick (approved):

PR #362, opened 2022-08-08
Implement remote tap support

### LINKED ISSUE #342: Ability to configure Taps via the API
This ticket is to investigate an approach to enable configuration of pktvisor taps via the pktvisor API. This API could then be leveraged by the Orb agent and Orb control plane to do tap configuration in the UI.

leoparente:


leoparente:
unit tests with code coverage check

rboucher-me (approved):
Looks good to me.

PR #368, opened 2022-08-16
GeoIP ASN prefix filtering

leoparente:


leoparente:
@rboucher-ns1 I did as `exactly match.`.. it would be costly to do as case insensitive... only if we cache the lower or upper case after database lookup...
What do you think?

I was able to test it with local pcap file. it seems to be working properly

src/handlers/net/NetStreamHandler.cpp
```diff
@@ -65,6 +65,20 @@ void NetStreamHandler::start()
...
     }
 
+    if (config_exists("only_geoloc_prefix")) {
+        _f_enabled.set(Filters::GeoLocPrefix);
+        for (const auto &prefix : config_get<StringList>("only_geoloc_prefix")) {
+            _f_geoloc_prefix.push_back(prefix);
+        }
+    }
+
+    if (config_exists("only_asn_prefix")) {
```
rboucher-me: For ASN, it's not really a "prefix" (though I see your point that it comes back as a string that includes both the ASN number and an ASN name.

In our case, we only want to filter based on the ASN number (exact match). We probably want to rename "only_asn_prefix" to something like "only_asn_networks" or "only_asn_numbers" or "only_asns"

rboucher-me (commented):
I'm good with case sensitive matching.

src/handlers/net/NetStreamHandler.cpp
leoparente: `only_asn_number` i think is better.

leoparente: We have been using singular on filter names: "`only_qname_suffix`" for instance

leoparente: Done

rboucher-me (approved):

PR #369, opened 2022-08-16
[feat] add automatic release generate

etaques:
This PR add automatic release generate with changelog compared from last created tag


manrodrigues (approved):
LGTM

PR #371, opened 2022-08-16
Implement Handler Event Proxy Concept

### LINKED ISSUE #370: Create HandlerEventProxy
HandlerEventProxy allow pass event proxy to other handlers in chaining mode. This allow reuse of Open Proxies for custom Handlers

leoparente:


weyrick (changes requested):
as discussed in slack, refactor for single input event proxy

leoparente:
> as discussed in slack, refactor for single input event proxy

Done! The only problem with this new approach is that it does not do a validation if the chaining will work like before... Should we add a logic for That?

like, before if we did [Dns, DHCP] it would return an error saying that DHCP could not attach to DNS Handler, but now as the inputProxy is PCAP, it will perform the chaining normally, but not data will be shown on DHCP module...
But Thinking on Handler Reuse and custom Handlers... This new approach is better

weyrick (approved):
This PR is great because it removes more than it adds :)

PR #373, opened 2022-08-18
Implement network interface scan and pick default tap

### LINKED ISSUE #152: setup default taps at startup if none specified

leoparente:


leoparente:
For Mac OS I was able only to validate the build, not the feature itself.
Tested in my local environment (Linux) and it seems to be working properly.

leoparente:
@weyrick need you to build and test this feature on MAC OS.

cmd/pktvisord/NetworkInterfaceScan.h
```diff
@@ -0,0 +1,51 @@
...
+            continue;
+        }
+#if defined(AF_PACKET)
+        if (int family = ifa->ifa_addr->sa_family; family == AF_PACKET && ifa->ifa_data != nullptr) {
+#elif defined(AF_LINK)
+        if (int family = ifa->ifa_addr->sa_family; family == AF_LINK && ifa->ifa_data != nullptr) {
+#else
+        if (false) {
+#endif
+            uint32_t *stats = reinterpret_cast<uint32_t *>(ifa->ifa_data);
```
weyrick: I think we should follow the man pages and use the actual stat struct rather than trying to interpret it directly like this. For example ```struct rtnl_link_stats *stats = ifa->ifa_data;``` from https://man7.org/linux/man-pages/man3/getifaddrs.3.html

The OSX version says ```     The ifa_data field references address family specific data.  For AF_LINK addresses it contains a pointer to the struct if_data (as defined in include file <net/if.h>) which contains various interface
     attributes and statistics.  For all other address families, it contains a pointer to the struct ifa_data (as defined in include file <net/if.h>) which contains per-address interface statistics.```

cmd/pktvisord/main.cpp
```diff
@@ -571,7 +572,17 @@ int main(int argc, char *argv[])
         exit(EXIT_FAILURE);
     }
 
+    std::string iface;
     if (args["IFACE"]) {
+        iface = args["IFACE"].asString();
+    } else {
```
weyrick: so I don't think we should automatically create a pcap policy if no arguments are passed, because by default i want the user to specify --admin-api or an interface still (which is only for legacy reasons). so, i think we should only do this interface selection and the legacy pcap policy if they pass `auto` as the interface.

however - i do think we can automatically create `default_pcap` tap on the most used interface _if_ there is no `default_pcap` in the config file.

weyrick (changes requested):

cmd/pktvisord/NetworkInterfaceScan.h
leoparente: I did that in that way because MacOS does not have `rtnl_link_stats` struct, and net/if does not have an interface to proper parser tx and rx packets, thats why I would like you to test this PR on MacOS so we know if it does work as expected or Not.

weyrick: Right but it contains a different structure which does have statistics. So that part would have to be in an #ifdef for each platform

leoparente: I've changed that by platform and added rtnl_link_stats for linux

cmd/pktvisord/main.cpp
leoparente: But we should lower the barrier Right? be able to execute pktvisord without any value...
The implementation was wrong before because for any case the default policy was being created. I've changed it to:
 It should only should find and create interface if no interface was specified AND config file was not set (orb agent always set a config file for instance)
 
 What do you think? Pr is updated with the approach that I mentioned

weyrick (approved):

PR #378, opened 2022-08-19
[FIX] remove clean due error on pipeline

etaques:
This PR remove clean from integration tests because is unnecessary on current approach


leoparente (approved):
LGTM

PR #387, opened 2022-08-24
[FIX] develog branch concurrency control

etaques:
This PR add commit hash on EC2 security group to control concurrency, permitting we have more than one pipeline running at same time on develop branch


manrodrigues (approved):
LGTM

PR #389, opened 2022-08-24
Add TopPorts group do DnsHandler

### LINKED ISSUE #382: Reduce dns port cardinality
The DNS handler contains a `top_udp_ports` metric, enabled by default, which is a top 10 DNS client ports table, intended to help identify certain traffic abuses (for example, a flood all coming from the same client source port). However, when there is no malicious traffic, the label is very high cardinality because this field in DNS is intentionally random. Analysis shows this cardinality can account for >40% of total metric series.

We can greatly reduced cardinality here by introducing a threshold to be included in the table output - for example, only include ports that have received more than 120 queries in a time bucket (a number we've researched from empirical data). we can also make this number configurable via handler config. This allows abuses to still be understood while minimizing the number of active time series.

leoparente:


src/Metrics.h
```diff
@@ -235,12 +237,27 @@ class TopN final : public Metric
...
+    double _percentile;
+
+    uint64_t _get_threshold() const
+    {
+        auto items = _fi.get_frequent_items(datasketches::frequent_items_error_type::NO_FALSE_NEGATIVES);
+        datasketches::kll_sketch<uint64_t> quantile;
+        for (uint64_t i = 0; i < std::min(_top_count, items.size()); i++) {
+            quantile.update(items[i].get_estimate());
+        }
+        if (quantile.is_empty()) {
```
leoparente, line 248: get_quantile throws it quantile is empty. that's why I needed to add this validation

src/Metrics.h
```diff
@@ -219,6 +219,8 @@ class Quantile final : public Metric
 template <typename T>
 class TopN final : public Metric
 {
+    static constexpr uint64_t KLL_DEFAULT = 0;
```
weyrick: let's name `DEFAULT_PERCENTILE_THRESHOLD`

src/Metrics.h
```diff
@@ -235,12 +237,27 @@ class TopN final : public Metric
     datasketches::frequent_items_sketch<T> _fi;
     size_t _top_count = 10;
     std::string _item_key;
+    double _percentile;
```
weyrick: `_percentile_threshold`

src/Metrics.h
```diff
@@ -235,12 +237,27 @@ class TopN final : public Metric
...
+
+    uint64_t _get_threshold() const
+    {
+        auto items = _fi.get_frequent_items(datasketches::frequent_items_error_type::NO_FALSE_NEGATIVES);
```
weyrick: let's pass this in - see below

src/Metrics.h
```diff
@@ -278,9 +295,12 @@ class TopN final : public Metric
     {
         auto section = json::array();
         auto items = _fi.get_frequent_items(datasketches::frequent_items_error_type::NO_FALSE_NEGATIVES);
+        auto threshold = _get_threshold();
```
weyrick: We call `get_frequent_items` here, and also in `_get_threshold`. I think it's a relatively expensive call so let's pass it into `_get_threshold` instead of calling it twice

src/Metrics.h
```diff
@@ -278,9 +295,12 @@ class TopN final : public Metric
...
         for (uint64_t i = 0; i < std::min(_top_count, items.size()); i++) {
-            section[i]["name"] = formatter(items[i].get_item());
-            section[i]["estimate"] = items[i].get_estimate();
+            if (items[i].get_estimate() >= threshold) {
+                section[i]["name"] = formatter(items[i].get_item());
+                section[i]["estimate"] = items[i].get_estimate();
+            }
```
weyrick, line 312: since this is sorted, we can add an `else` here to break when estimate is < threshold

src/Metrics.h
```diff
@@ -235,12 +237,27 @@ class TopN final : public Metric
...
+        return quantile.get_quantile(_percentile);
+    }
 
 public:
-    TopN(std::string schema_key, std::string item_key, std::initializer_list<std::string> names, std::string desc)
+    TopN(std::string schema_key, std::string item_key, std::initializer_list<std::string> names, std::string desc, uint64_t percentile = KLL_DEFAULT)
         : Metric(schema_key, names, std::move(desc))
         , _fi(MAX_FI_MAP_SIZE, START_FI_MAP_SIZE)
         , _item_key(item_key)
+        , _percentile(static_cast<double>(percentile) / 100)
```
weyrick: we probably want to add a check here to make sure the percentile is between 0 and 1.0 and throw if not

weyrick (changes requested):

src/Metrics.h
leoparente: done

src/Metrics.h
leoparente: done

src/Metrics.h
leoparente: ok. done

src/Metrics.h
leoparente: done

src/Metrics.h
leoparente: done

src/Metrics.h
leoparente: done

leoparente:
Things on Metrics contructor is a problem due the way or architecture for MetricsBucket is implemented... E should probably do a way like we did for topn count implementation: https://github.com/ns1labs/pktvisor/pull/308/files

src/Metrics.h
```diff
@@ -235,13 +237,30 @@ class TopN final : public Metric
     datasketches::frequent_items_sketch<T> _fi;
     size_t _top_count = 10;
     std::string _item_key;
+    double _percentile_threshold;
+
+    uint64_t _get_threshold(std::vector<typename datasketches::frequent_items_sketch<T>::row> &items) const
```
weyrick: items can be const?

weyrick (approved):

weyrick:
> Things on Metrics contructor is a problem due the way or architecture for MetricsBucket is implemented... E should probably do a way like we did for topn count implementation: https://github.com/ns1labs/pktvisor/pull/308/files

not sure what you mean?

src/Metrics.h
leoparente: yes, I will change that

weyrick (approved):

PR #393, opened 2022-08-27
separate DnsHandler into DnsWireHandler and DnsXactHandler

### LINKED ISSUE #341: DNS Handler - Create DnsHandler V2
The DNS handler should still be able to provide `xact` metrics when an `rcode` filter is applied.

Currently, if a filter is applied such as `"only_rcode": 3` (only NXDOMAIN), the `xact` metrics are empty (all zeroes).

leoparente:


leoparente:
We will genarate a version 2 of dns handler

PR #397, opened 2022-08-29
only enable pcap debug if the value is True

### LINKED ISSUE #395: only apply debug to PCAP input if it is true
Currently, if the configuration exists, this is enough to activate the debug. The ideal would be that only if the variable was (insensitive case) true the parameter was activated.

https://github.com/ns1labs/pktvisor/blob/develop/src/inputs/pcap/PcapInputStream.cpp#L101

leoparente:


manrodrigues (approved):
LGTM

PR #405, opened 2022-08-30
API return codes

### LINKED ISSUE #246: HTTP response status code for policies and taps creation must be 201
Pktvisor current response status code for policies and taps creation is 200, but must be 201

image

### LINKED ISSUE #247: HTTP response status code for policies and taps removal must be 204
Once there is no message or information on response, status code expected for policy and taps removal is 204, but currently pktvisor return 200. <br><br>

image

PS:
I'm based on it:

### LINKED ISSUE #399: Set BPF filter to " " if pcap_file exists and no BPF filter is defined
For now, if a pcap file exists but no bpf filter is configured, the policy will fail (https://github.com/ns1labs/pktvisor/blob/develop/src/inputs/pcap/PcapInputStream.cpp#L93). However, if the user configure bpf as " ", it will work. We can thus implement this logic transparently to the user, in the same way as done here: https://github.com/ns1labs/pktvisor/blob/develop/src/inputs/pcap/PcapInputStream.cpp#L133.

image

### LINKED ISSUE #400: HTTP response status code for policies and taps conflict name must be 409
When a policy or a tap is created with the same name as one already existent, the status code returned must be 409 (conflict), instead of 500(internal server error).. It's not useful for the users that all the error returns the same code.

leoparente:
`Orb Agent `was already changed to accept success from 200 to 299

### DIFF
--- a/src/CoreServer.cpp
+++ b/src/CoreServer.cpp
@@ -242,9 +242,10 @@ void CoreServer::_setup_routes(const PrometheusConfig &prom_config)
             for (auto &mod : taps) {
                 mod->info_json(j[mod->name()]);
             }
+            res.status = 201;
             res.set_content(j.dump(), "text/json");
         } catch (const std::exception &e) {
-            res.status = 500;
+            (std::string(e.what()).find("already") != std::string::npos) ? res.status = 409 : res.status = 500;
             j["error"] = e.what();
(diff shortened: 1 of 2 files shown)


manrodrigues (approved):
LGTM

PR #409, opened 2022-08-31
[FIX] paralelize integration tests

etaques:
Paralelize integration tests

### DIFF
--- a/.github/workflows/build-develop.yml
+++ b/.github/workflows/build-develop.yml
@@ -330,7 +330,7 @@ jobs:
           pip3 install --upgrade pip
           pip3 install -r requirements.txt
           pip3 install -r requirements.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/build-develop.yml
```diff
@@ -330,7 +330,7 @@ jobs:
           pip3 install --upgrade pip
           pip3 install -r requirements.txt
           pip3 install -r requirements.txt
-          behavex -t=@smoke
+          behavex -t=@smoke --parallel-processes=4 --parallel-scheme=scenario
```
manrodrigues: why only 4 processes? i guess we can use more.. i'm running locally with 20.. but you can put at least 10.. to make it faster..

etaques: Ok great. So I will change :)

leoparente (approved):

PR #426, opened 2022-09-05
Implement merge_equal feature on Policies

leoparente:


src/AbstractMetricsManager.h
```diff
@@ -519,6 +519,26 @@ class AbstractMetricsManager
         _metric_buckets.at(period)->to_prometheus(out, add_labels);
     }
 
+    void window_external_prometheus(std::stringstream &out, AbstractMetricsBucket *bucket, Metric::LabelMap add_labels = {}) const
+    {
+        if (auto custom = dynamic_cast<MetricsBucketClass *>(bucket); custom) {
+            custom->to_prometheus(out, add_labels);
+        } else {
+            throw std::runtime_error("invalid bucket type");
```
weyrick: this could that happen from a policy right, only programmer error?

src/Policies.cpp
```diff
@@ -286,6 +297,10 @@ void Policy::start()
     }
     assert(_input_streams.size());
     spdlog::get("visor")->info("policy [{}]: starting", _name);
+
+    // configurations
+    _merge_equal = (config_exists("merge_equal") && config_get<bool>("merge_equal"));
```
weyrick: how about "merge_like_handlers" instead of "merge_equal"

weyrick (changes requested):
Looks pretty good! Let's get some units and update to the RFC as well.

src/AbstractMetricsManager.h
leoparente: I've changed that to static and added a comment. It only programmer error

src/Policies.cpp
leoparente: cool. changed that

weyrick (approved):
cool let's try it!

PR #432, opened 2022-09-08
Extract geodb on pktvisor image

etaques:
This PR include on container entrypoint geodb extraction


leoparente (changes requested):
We should check if user provided its own database before extracting it

PR #439, opened 2022-09-09
Track Request Ack transactions on DHCP Handler

leoparente:


etaques (approved):
LATM

src/handlers/dhcp/DhcpStreamHandler.h
```diff
@@ -24,6 +25,8 @@ class DhcpMetricsBucket final : public visor::AbstractMetricsBucket
 protected:
     mutable std::shared_mutex _mutex;
 
+    TopN<std::string> _dhcp_clients;
```
weyrick: Should have "top" in the name... _top_clients or _top_dhcp_clients

src/handlers/dhcp/DhcpStreamHandler.cpp
```diff
@@ -177,15 +183,43 @@ void DhcpMetricsBucket::process_dhcp_layer(bool deep, pcpp::DhcpLayer *payload,
...
+{
+    if (!deep) {
+        return;
+    }
+    // lock for write
+    std::unique_lock lock(_mutex);
 
-void DhcpMetricsManager::process_dhcp_layer(pcpp::DhcpLayer *payload, [[maybe_unused]] PacketDirection dir, pcpp::ProtocolType l3, pcpp::ProtocolType l4, [[maybe_unused]] uint32_t flowkey, uint16_t src_port, uint16_t dst_port, timespec stamp)
+    if (auto client_ip = payload->getYourIpAddress(); client_ip.isValid()) {
+        _dhcp_clients.update(client_ip.toString() + "/" + xact.mac_address + "/" + xact.hostname);
```
weyrick: hm do we want/need the ip here? it seems to me mac and hostname actually identify the client. otherwise we will get a new entry if the same mac/host gets a different IP address?

weyrick (changes requested):

src/handlers/dhcp/DhcpStreamHandler.cpp
leoparente: I've changed the order of parameters to MAC/HOST/IP

leoparente: done

weyrick (approved):

PR #444, opened 2022-09-12
Add Aggregate to Merge quantile and rates functions

### LINKED ISSUE #437: Proper quantile merge when Merging like handlers

leoparente:


src/AbstractMetricsManager.h
```diff
@@ -65,7 +65,7 @@ class AbstractMetricsBucket
     const std::bitset<GROUP_SIZE> *_groups;
 
     // merge the metrics of the specialized metric bucket
-    virtual void specialized_merge(const AbstractMetricsBucket &other) = 0;
+    virtual void specialized_merge(const AbstractMetricsBucket &other, bool aggregate) = 0;
```
weyrick: Rather than `bool aggregate`, let's add a new enum class with DEFAULT and SUM items, and have the parameter be `agg_operator` defaulting to DEFAULT. for the quantile merge we will pass SUM

src/Metrics.h
```diff
@@ -154,9 +154,23 @@ class Quantile final : public Metric
...
-    void merge(const Quantile &other)
+    void merge(const Quantile &other, bool aggregate)
     {
-        _quantile.merge(other._quantile);
+        if (aggregate && !_quantile.is_empty()) {
+            if (other._quantile.is_empty()) {
+                return;
+            }
+            const uint32_t num_quantiles = 10;
+            auto quantiles = _quantile.get_quantiles(num_quantiles);
```
weyrick: I'm not sure what passing 10 here does? Why wouldn't we want to use the same quantiles we use in output - 50, 90, 95, 99?

weyrick (changes requested):

src/Metrics.h
leoparente: I thought I little about this...
I'm generating a new quantile variable with the sum of current and other quantiles. The problem of only getting 4 values would be the displayed below:

image

src/AbstractMetricsManager.h
leoparente: Great, Ok

src/Metrics.h
weyrick: Ah right, the output of a sum aggregation across quantiles actually isn't another quantile - it's a vector of individual results at each percentile (so like, an array of four numbers). Would that work in this design?

leoparente: You are right, That's why I got the 10 quantiles for each and sum them... so we have 10 values that were sum... So when we output the 0.5 0.9 0.95 0.99 it would be a more precise value 

The return of this merge is a quantile structure with 10 data point in which each data point is a sum of quantile the merge quantiles

so:
new_quantile[0] = q0.1 + other_q0.1
new_quantile[1] = q0.2 + other_q0.2
....
new_quantile[9] = q1.0 + other_q1.0

leoparente: proper add the sum values

weyrick (approved):
cool

PR #450, opened 2022-09-13
Create Visor DNS lib

leoparente:


libs/visor_dns/DnsAdditionalRecord.h
```diff
@@ -9,7 +9,7 @@
 #include <algorithm>
 #include <array>
 
-namespace visor::handler::dns {
+namespace visor::dns {
```
weyrick: let's do `visor::lib::dns`

weyrick (approved):

libs/visor_dns/DnsAdditionalRecord.h
leoparente: done

weyrick (approved):

PR #458, opened 2022-09-17
Load new GeoDB reference to Handler Plugins on init

leoparente:


src/HandlerModulePlugin.h
```diff
@@ -17,8 +17,10 @@ class InputEventProxy;
 
 class HandlerModulePlugin : public AbstractPlugin
 {
-
 public:
+    static geo::MaxmindDB *asn;
```
weyrick, line 21: can/should we make these `const`? default to `nullptr`?

weyrick (approved):

src/HandlerModulePlugin.h
leoparente: it cannot be const because we should change it reference/value on runtime.


The default is a valid reference
HandlerModulePlugin.cpp 
```
geo::MaxmindDB *HandlerModulePlugin::asn = &geo::GeoASN();
geo::MaxmindDB *HandlerModulePlugin::city = &geo::GeoIP();
```

For static module this reference is enough... but for custom runtime plugins this reference exists but will be always not enabled DBs.. then when we init the custom handler plugin module we pass the correct loaded reference;

leoparente: So asn and city are never nullptr

PR #462, opened 2022-09-21
Flow Metrics Enrichment

leoparente:


src/handlers/flow/FlowStreamHandler.cpp
```diff
@@ -47,6 +67,44 @@ void FlowStreamHandler::start()
...
+                if (data.size() > 4) {
+                    device->interfaces[if_index] = InterfaceEnrich{data[2], data[4]};
+                } else {
+                    device->interfaces[if_index] = InterfaceEnrich{data[2], std::string()};
+                }
+            }
+        }
+    }
+
+    if (config_exists("concatenate_interface")) {
```
leoparente, line 104: @rboucher-ns1 the idea here is to pass like the data that you want to concatenate the `deviceId`
like 
```
config:
   concatenate_interface: "eth0"
```

Then, based on that we generate the label `interface="1.2.3.4.5|eth0` and if it has enrichment it would be `interface="route1|eth0`

leoparente: Would like your opinion oh that name `concatenate_interface`

PR #466, opened 2022-09-22
Allow no metrics handler configuration

leoparente:
- This allow using a chaining handler only to filter data


leoparente:
changed to approach on PR https://github.com/ns1labs/pktvisor/pull/467

PR #467, opened 2022-09-23
Add default 'all' handlers metric group

leoparente:


RFCs/2021-04-16-76-collection-policies.md
```diff
@@ -109,6 +109,10 @@ visor:
           # Chaning handlers example. It needs proper indentation as shown below
           - upstream_dns:
               type: dns
+              metric_groups:
+                # disable all metric groups - no metrics will be scrapped for this handler
```
weyrick: typo 'scraped'. let's also document that you can both enable or disable all

weyrick (approved):
looks good just docs

RFCs/2021-04-16-76-collection-policies.md
leoparente: done

PR #470, opened 2022-09-24
Implement LatLong data on GeoDB city lookup

### LINKED ISSUE #112: Support lat/long and/or geohash in top geoloc
Currently the top N geo locations include only a "name" indicating region/country/city. It would be desirable also include lat/long and/or geohash of the location so that it could be used by dashboards to map locations on a world view, e.g. Grafana world map plugin.

In JSON output this would live alongside "name" and "estimate" keys. In Prometheus output these would be new labels.

leoparente:
- Add label  "lat_long" for prometheus output
- Add new key "lat_long" for json output

### DIFF
--- a/src/GeoDB.h
+++ b/src/GeoDB.h
@@ -10,12 +10,31 @@
 #pragma GCC diagnostic pop
 #include <memory>
 #include <mutex>
+#include <nlohmann/json.hpp>
+#include <ostream>
 #include <shared_mutex>
 #include <string>
+#include <variant>
 
 #include "VisorLRUList.h"
 
 namespace visor::geo {
+
+struct City {
+    std::string location;
+    std::string latitude;
+    std::string longitude;
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE(City, location);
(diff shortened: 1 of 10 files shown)


src/GeoDB.h
```diff
@@ -33,10 +52,10 @@ class MaxmindDB
     /*
      * These routines accept both IPv4 and IPv6
      */
-    std::string getGeoLocString(const char *ip_address) const;
-    std::string getGeoLocString(const struct sockaddr *sa) const;
-    std::string getGeoLocString(const struct sockaddr_in *sa4) const;
-    std::string getGeoLocString(const struct sockaddr_in6 *sa6) const;
+    City getGeoLocString(const char *ip_address) const;
```
weyrick: let's rename to `getGeoLoc`

weyrick (changes requested):

src/GeoDB.h
leoparente: done

weyrick (approved):

PR #471, opened 2022-09-24
Update InputStream iface config when using auto

### LINKED ISSUE #436: Expose the interface observed when "auto" is chosen as iface
@manrodrigues commented on [Fri Sep 09 2022](https://github.com/ns1labs/ns1-pktvisor/issues/37)

When "auto" is chosen as iface, pktvisor binds on the most used interface, but there is no way to the user knows which iface is being observed and as a user I want to know it.

leoparente:


manrodrigues (approved):
LGTM

PR #472, opened 2022-09-24
Add version "1.0" as default for Tap and Policy creation

### LINKED ISSUE #406: Policies: insert version "1.0" as default
The policies structure requires the version to be filled in. This version indicates the schema version expected in the policy. To make the policy creation experience easier, we can set version "1.0" (the only one available so far) by default. Thus, if the field is not set, it will have the value 1.0. If the user passes another value, the value passed will be used.

Expected behavior:

POST request on endpoint `http://localhost:{port}/api/v1/policies` must support `version:` as a configurable field, but it must NOT be required. If it not exist on body request, policy must be created with `version: "1.0"`

leoparente:


manrodrigues (approved):
LGTM

PR #473, opened 2022-09-26
add top no error qnames on dns handler

### LINKED ISSUE #468: Dns Top NOERROR qnames

leoparente:


leoparente:
@weyrick @rboucher-ns1 I think that new metric would have a high cardinality... wouldn't that be a problem?

rboucher-me:
> @weyrick @rboucher-ns1 I think that new metric would have a high cardinality... wouldn't that be a problem?

I think the cardinality will be very dependent on the DNS traffic profile: it might be high in certain cases, and modest in others. We'll need to look if this overlaps significantly with the qname2 and qname3  top metrics, and see if we need to keep all three.

rboucher-me (approved):
Looks good to me.

weyrick:
What about putting this in a metric group and turning off by default? I'm definitely reluctant to add this because of cardinality, overlap, and the fact that it's a "shortcut" table (i.e. something you can get by creating the right policy with a filter ... right?)

leoparente:
> What about putting this in a metric group and turning off by default? I'm definitely reluctant to add this because of cardinality, overlap, and the fact that it's a "shortcut" table (i.e. something you can get by creating the right policy with a filter ... right?)

I'm ok to create a new group. but what would be the name for that group?

rboucher-me:
I think it makes sense to have this disabled by default. We will have to revisit these metrics as part of the DNS Handler v2 work, imo.

weyrick:
Ok new group could be "extra top tables" or something? Or just "beta group"?

rboucher-me:
Good question. Making it part of `TopQnames` wouldn't help much... ;)

Would `TopQnameDetails` make sense? (arguably, `topSizedQnameResp` would probably make sense to be in that group too)

leoparente:
@rboucher-ns1 @weyrick `TopQnameDetails` Then?

rboucher-me (approved):
Looks good.

PR #480, opened 2022-09-30
Create IpPort struct with port name enrichment

leoparente:


src/handlers/flow/tests/test_flows.cpp
```diff
@@ -44,7 +44,11 @@ TEST_CASE("Parse sflow stream", "[sflow][flow]")
     CHECK(j["devices"]["192.168.0.13"]["top_src_ips_packets"][0]["estimate"] == 71180000);
     CHECK(j["devices"]["192.168.0.13"]["top_src_ips_packets"][0]["name"] == "10.4.1.2");
     CHECK(j["devices"]["192.168.0.13"]["top_dst_ports_bytes"][0]["estimate"] == 170879120000);
+#if __APPLE__
```
leoparente, line 47: This is a proof that this implementation works. port 5001 is mapped by default on macOS and not on Linux =]

src/IpPort.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+namespace visor {
+
+std::ostream &operator<<(std::ostream &os, const visor::IpPort &p)
+{
+    os << std::to_string(p.port);
+    return os;
+}
+
+std::string IpPort::get_name() const
```
weyrick: please name `get_service` or `get_service_name` instead

src/IpPort.cpp
```diff
@@ -0,0 +1,31 @@
+#include "IpPort.h"
+
+namespace visor {
```
weyrick: how about a sub namespace for `network`? so this would be `visor::network::IpPort` ... we may move things here over time out of pcapplusplus.

weyrick (changes requested):
Dropped a few comments. But also (cc @rboucher-ns1) I think this needs to be optionally turned off because sometimes people put different things on common ports. Or do we plan to solve that use case with custom maps?

src/IpPort.cpp
leoparente: done

src/IpPort.cpp
leoparente: done

rboucher-me:
> Dropped a few comments. But also (cc @rboucher-ns1) I think this needs to be optionally turned off because sometimes people put different things on common ports. Or do we plan to solve that use case with custom maps?

We need to be able to support customization of the port name mappings, though not necessarily immediately. Ideally this approach would allow us to support this with minimal effort.

weyrick (approved):

PR #489, opened 2022-10-04
Create 2022-10-04-78-probe-policies.md

rboucher-me:
Here's a first draft of what a `probe` policy might look like, and what it should include.


rboucher-me:
After discussion with Shannon, this is no longer required

PR #491, opened 2022-10-05
Create Net Probe input

### LINKED ISSUE #487: Spike: Ping Probe Implementation
- [x] Be able to create a Probe policy on pktvisor
- [x] Be able to ping ipv4 addresses

### LINKED ISSUE #490: Sample configurations for "netprobe" tap and policy
The purpose of this ticket is to discuss how to structure the new `netprobe` tap and handler. Here are a few concepts to keep in mind in reviewing this:
-  the `netprobe` _tap_ provides the basic facilities to run network tests, but should only include host specific configuration settings and default configuration overrides
- the `netprobe` _input_ is where the network tests are defined and configured, as well as any default configuration overrides
- the `netprobe` _handler_ is where the metrics that are to be measured and collected are configured

Here is a sample configuration for the `netprobe` tap:
```yaml
version: "1.0"

visor:
  taps:
    default_netprobe:
      input_type: netprobe
      config:
        maximum_concurrent_tests: 10
        ip_source_binding: 127.0.0.1
      tags:
        virtual: true
        vhost: 1 
```

Here is a sample policy for the `netprobe` input and handler:
```yaml
version: "1.0"

visor:
  policies:
    basic_ping_policy:
      kind: collection
      description: "basic PING netprobe policy"
      input:
        tap: default_netprobe
        input_type: netprobe
        config:
          test_type: ping
          interval_msec: 2000
          timeout_msec: 1000
          packets_per_test: 10
          packets_interval_msec: 25
          packet_payload_size: 56
          disable_scout_packet: false
          disable_integrity_check: false
          targets:
            test_1_name:
              target: foo.bar
            test_2_name:
              target: 10.0.0.1
              tos: EF
      handlers:
        config:
          num_periods: 2 #default is 5
        modules:
          default_ping:
            type: netprobe
            config:
              latency_units: msec
            metric_groups:
              enable:
                - quantiles
                - dns_resolution
              disable:
                - jitter
```

leoparente:


weyrick (commented):
looks like good direction!

src/handlers/netprobe/README.md
```diff
@@ -0,0 +1,7 @@
+# Net Probe Metrics Stream Handler
+
+This directory contains the Net Probe stream handler
+
+It can attach to pcap input streams to summarize Net Probe traffic
```
weyrick: let's make this real before we forget :)

src/inputs/netprobe/NetProbeInputStream.h
```diff
@@ -0,0 +1,89 @@
...
+
+    void probe_fail_cb(ErrorType e, TestType t, const std::string &n)
+    {
+        probe_fail_signal(e, t, n);
+    }
+
+    // handler functionality
+    // IF THIS changes, see consumer_count()
+    // note: these are mutable because consumer_count() calls slot_count() which is not const (unclear if it could/should be)
+    mutable sigslot::signal<pcpp::Packet &, TestType, const std::string &> probe_recv_signal;
```
weyrick: i think this won't work for e.g. HTTP test but don't have to worry about that here

weyrick (commented):
cool i like the layout of the classes, good direction 👍

src/handlers/netprobe/README.md
leoparente: done

src/inputs/netprobe/NetProbeInputStream.cpp
```diff
@@ -0,0 +1,232 @@
...
+    _timer->on<uvw::ErrorEvent>([this](const auto &err, auto &handle) {
+        _logger->error("[{}] TimerEvent error: {}", _name, err.what());
+        handle.close();
+    });
+
+    for (const auto &ip : _ip_list) {
+        std::unique_ptr<NetProbe> probe{nullptr};
+        if (_type == TestType::Ping) {
+            probe = std::make_unique<PingProbe>(_id, ip.first);
+        }
```
weyrick, line 169: need `else` here or the following probe-> will segfault. same in dns section below

src/inputs/netprobe/PingProbe.h
```diff
@@ -0,0 +1,83 @@
...
+#include <mutex>
+#include <optional>
+#include <sigslot/signal.hpp>
+#include <uvw/async.h>
+#include <uvw/poll.h>
+#include <uvw/timer.h>
+
+namespace visor::input::netprobe {
+
+class PingReceiver
```
weyrick, line 39: please document the threading strategy here: that there is one of these per pktvisor process (right?)

src/inputs/netprobe/PingProbe.h
```diff
@@ -0,0 +1,83 @@
...
+    void _setup_receiver();
+
+public:
+    static sigslot::signal<pcpp::Packet &, timespec> recv_signal;
+
+    PingReceiver();
+    ~PingReceiver();
+};
+
+class PingProbe final : public NetProbe
```
weyrick, line 65: and document here too: one PingProbe per target but shared socket, one per pktvisor process?

src/inputs/netprobe/PingProbe.cpp
```diff
@@ -0,0 +1,314 @@
...
+        _recv_connection = PingReceiver::recv_signal.connect([this](pcpp::Packet &packet, timespec stamp) { _recv(packet, TestType::Ping, _name, stamp); });
+
+        (_sequence == USHRT_MAX) ? _sequence = 0 : _sequence++;
+        _send_icmp_v4(_internal_sequence);
+        _internal_sequence++;
+        _timeout_timer->start(uvw::TimerHandle::Time{_timeout_msec}, uvw::TimerHandle::Time{0});
+        _internal_timer->start(uvw::TimerHandle::Time{_packets_interval_msec}, uvw::TimerHandle::Time{_packets_interval_msec});
+    });
+
+    ++_sock_count;
```
weyrick: let's expose this (and maybe other internals) in the `info_json` endpoint?

src/handlers/netprobe/NetProbeStreamHandler.h
```diff
@@ -0,0 +1,118 @@
...
+
+struct Target {
+    Quantile<uint64_t> time_us;
+    Counter attempts;
+    Counter successes;
+    Counter minimum;
+    Counter maximum;
+
+    Target()
+        : time_us(NET_PROBE_SCHEMA, {"response_quantiles_us"}, "Net Probe quantile in microseconds")
```
weyrick, line 30: so the plan is to have these metrics across all probe types? cc @rboucher-ns1

src/handlers/netprobe/NetProbeStreamHandler.cpp
```diff
@@ -0,0 +1,208 @@
...
+}
+
+void NetProbeMetricsBucket::to_prometheus(std::stringstream &out, Metric::LabelMap add_labels) const
+{
+    std::shared_lock r_lock(_mutex);
+
+    for (const auto &target : _targets_metrics) {
+        auto target_labels = add_labels;
+        auto targetId = target.first;
+        target_labels["target"] = targetId;
```
weyrick, line 105: is targetId always an IP, or will it be a hostname if they specified it that way in the policy?

weyrick (changes requested):
looking great!

src/inputs/netprobe/NetProbeInputStream.cpp
leoparente: done

src/handlers/netprobe/NetProbeStreamHandler.cpp
leoparente: actually it will be the specified name. specified on the policy
```
....
          targets:
            test_1_name:
              target: foo.bar
            test_2_name:
              target: 10.0.0.1
              tos: EF
```


should we have two labels? one for target name and other for the target? If so... how should we show that on `json` return

src/inputs/netprobe/PingProbe.h
leoparente: one per thread and not process in this case.

leoparente: added doxygen

src/inputs/netprobe/PingProbe.cpp
leoparente: done

src/handlers/netprobe/NetProbeStreamHandler.cpp
weyrick: ah interesting. ok i think it's ok to land it this way - this interface can be considered beta for now anyway. but it provides a way to keep the label consistent in the dashboards, which is what I was concerned about 👍

src/handlers/netprobe/NetProbeStreamHandler.h
weyrick: we can refine post land too (we'll be in beta)

weyrick (approved):

PR #496, opened 2022-10-12
Update Flow Handler

### LINKED ISSUE #492: Flow Handler Updates
The following Flow handler updates should be considered. Please refer to this [doc](https://github.com/ns1labs/orb/wiki/Network-Traffic-Observability-using-Flow) for more details.

1. Deprecating `first_filter_if_as_label: true`
2. Implementing `enrichment: false #default should be true`
3. Investigate implementing `mute_dynamic_ports: true #default should be true`
This setting would anonymize (or "mute") dynamic port numbers (49152 to 65535) to reduce metric cardinality (for what are essentially random port numbers). This should apply to all metrics that use port numbers (whether enriched or not).
4. Implementing single device and interface filters:
```yaml
      filter:
        device: 192.243.6.193
        interface: 6
```
5. Restructuring metric groups along the lines of these proposed groups:
```
	ByPackets,
	ByBytes,
	Counters,
	TopPorts,
	TopIPs,
	TopIP_Ports,
	TopGeo,
	Cardinality,
	TopConversations,
	TopInterfaces
```
6. Order processing `metric_groups` such that `disable` is processed before `enable`. This would then support this type of configuration:
```yaml
      metric_groups:
        enable:
          - interfaces
        disable:
          - all
```
7. Flow metrics should be structured in such a way as to always specify the direction (`in|out`) and the unit of measure (`bytes|packets`), such as:
`flow_(in|out)_tcp_(bytes|packets)` instead of `flow_tcp`
`flow_top_(in|out)_(dst|src)_port_(bytes|packets)` instead of `flow_top_(dst|src)_ports_(bytes|packets)`
`flow_top_(in|out)_(dst|src)_ip_(bytes|packets)` instead of `flow_top_(dst|src)_ip_(bytes|packets)`
`flow_top_(in|out)_(dst|src)_ip_port_(bytes|packets)` instead of `flow_top_(dst|src)_ip_port_(bytes|packets)`
8. Add new counter metrics for the specified device interface:
`flow_(in|out)_(bytes|packets)`

leoparente:
- Move metrics to Interface level
    - remove first_filter_if_as_label config
    - add enrichment (true/false) config
    - add metrics based on In and Out and not Total

### DIFF
--- a/src/handlers/flow/FlowStreamHandler.h
+++ b/src/handlers/flow/FlowStreamHandler.h
@@ -150,7 +150,7 @@ struct Counters {
         , OtherL4(FLOW_SCHEMA, {direction + "_other_l4_" + metric}, "Count of " + direction + " " + metric + " which are not UDP or TCP")
         , IPv4(FLOW_SCHEMA, {direction + "_ipv4_" + metric}, "Count of " + direction + " IPv4 by " + metric)
         , IPv6(FLOW_SCHEMA, {direction + "_ipv6_" + metric}, "Count of " + direction + " IPv6 by " + metric)
-        , total(FLOW_SCHEMA, {direction + "_" + metric}, "Count of total " + direction + " flows by " + metric)
+        , total(FLOW_SCHEMA, {direction + "_" + metric}, "Count of " + direction + " " + metric)
     {
     }
 };
--- a/src/handlers/flow/FlowStreamHandler.h
+++ b/src/handlers/flow/FlowStreamHandler.h
@@ -27,12 +27,16 @@ static constexpr const char *FLOW_SCHEMA{"flow"};
 
 namespace group {
 enum FlowMetrics : visor::MetricGroupIntType {
-    Counters,
+    ByPackets,
+    ByBytes,
     Cardinality,
     Conversations,
+    Counters,
+    TopPorts,
+    TopIPs,
+    TopIPPorts,
     TopGeo,
-    TopByPackets,
-    TopByBytes
+    TopInterfaces
 };
 }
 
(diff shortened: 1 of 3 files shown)


src/handlers/flow/FlowStreamHandler.h
```diff
@@ -215,60 +250,53 @@ class FlowMetricsBucket final : public visor::AbstractMetricsBucket
...
 
-    inline void set_enrich_data(std::unordered_map<std::string, std::string> *concat, EnrichMap *enrich_data)
+    inline void set_enrich_data(EnrichMap *enrich_data)
     {
-        _concat_if = concat;
         _enrich_data = enrich_data;
     }
 
-    inline void process_filtered(uint64_t filtered)
+    inline void process_filtered(uint64_t filtered, std::string device)
```
weyrick: string can be `const std::string&`

weyrick (approved):
other than that one note, code looks ok to me

src/handlers/flow/FlowStreamHandler.h
leoparente: done

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -78,87 +82,105 @@ struct FlowPacket {
...
+    Counter IPv4;
+    Counter IPv6;
+    Counter total;
+    Counters(std::string direction, std::string metric)
+        : UDP(FLOW_SCHEMA, {direction + "_udp_" + metric}, "Count of " + direction + " UDP by " + metric)
+        , TCP(FLOW_SCHEMA, {direction + "_tcp_" + metric}, "Count of " + direction + " TCP by " + metric)
+        , OtherL4(FLOW_SCHEMA, {direction + "_other_l4_" + metric}, "Count of " + direction + " " + metric + " which are not UDP or TCP")
+        , IPv4(FLOW_SCHEMA, {direction + "_ipv4_" + metric}, "Count of " + direction + " IPv4 by " + metric)
+        , IPv6(FLOW_SCHEMA, {direction + "_ipv6_" + metric}, "Count of " + direction + " IPv6 by " + metric)
+        , total(FLOW_SCHEMA, {direction + "_" + metric}, "Count of total " + direction + " flows by " + metric)
```
rboucher-me: Does this replace the previous `flow_records_total`? metric? Not sure we need the direction here, but if it comes for free... we can work with it

rboucher-me (approved):
LGTM

PR #519, opened 2022-11-02
Create Lib Transaction (single header)

leoparente:


src/handlers/dns/v2/DnsStreamHandler.h
```diff
@@ -4,14 +4,14 @@
 
 #pragma once
 
+#include "../../../../libs/visor_transaction/TransactionManager.h"
```
weyrick: we shouldn't need relative directories if the CMake is setup right, correct?

weyrick (approved):

src/handlers/dns/v2/DnsStreamHandler.h
leoparente: oh this was done automatically, I will fix that

PR #535, opened 2022-11-15
Add proper direction of DNSTAP packets

### LINKED ISSUE #524: Track direction with dnstap
Add hostspec direction to dnstap packets

leoparente:


leoparente:
@weyrick  Question: Should I do that direction track at Handler level or should be on DnstapInput and add direction parameter on `dnstap_signal`?

leoparente:
Just looking at code... it is better to perform at Input level like pcap does

weyrick (approved):

PR #536, opened 2022-11-16
Fail to start handler if it has a not supported/invalid configuration

### LINKED ISSUE #521: Insert warnings when some configuration or filter does not exist in pktvisor
I'm opening this card more as a suggestion than a request.

I think it would be very helpful for the user if pktvisor had some form of warning when a non-existent filter or configs is entered.

I understand that it is not appropriate to constrain and fail when something incorrect/non-existent is defined. But it's very anoying that this isn't even notified. I myself have spent a lot of time investigating the problem and the problem was that I had a typo when typing the filter name.

I think we could have some log of "this configuration/filter is not available and therefore will be disregarded"

leoparente:


src/StreamHandler.h
```diff
@@ -111,6 +119,23 @@ class StreamMetricsHandler : public StreamHandler
...
+            if (std::any_of(_window_config_defs.begin(), _window_config_defs.end(), [config](const auto &def) { return config == def; })) {
+                continue;
+            } else if (std::any_of(config_defs.begin(), config_defs.end(), [config](const auto &def) { return config == def; })) {
+                continue;
+            } else if (config == "enable") {
+                continue;
+            } else if (config == "disable") {
+                continue;
+            }
+            throw StreamHandlerException(fmt::format("{} is an invalid/unsupported config or filter. The valid configs/filters are: {}", config, fmt::join(config_defs, ", ")));
```
leoparente: @weyrick Should I return the window configs like `deep_sample_rate` here as well?

weyrick: hm, i guess so

weyrick (approved):

src/StreamHandler.h
leoparente: Ok, I will do that

PR #540, opened 2022-11-17
Add query and response filters and fix rcode filter on DnsHandler V1

leoparente:


src/handlers/dns/v1/tests/test_dns_layer.cpp
```diff
@@ -503,7 +503,7 @@ TEST_CASE("DNS Filters: answer_count", "[pcap][dns]")
 
     auto counters = dns_handler.metrics()->bucket(0)->counters();
 
-    CHECK(counters.UDP.value() == 16);
+    CHECK(counters.UDP.value() == 4);
```
leoparente, line 506: This shows that the queries where not being filtered out for rcode filter

rboucher-me (approved):
Looks good.

PR #551, opened 2022-11-24
Feature/csv port enrichment

### LINKED ISSUE #546: Summarization mechanism for `port` labels in Flow metrics
The `port` label associated with the `top_(in|out)_(src_dst)_ports_(bytes|packets)` flow metrics has very high cardinality, primarily because source client ports are randomly chosen (typically in the 49152–65535 range). Information on individual port numbers is of limited value, though aggregate measures could still provide value. This ticket proposes one approach to reduce the cardinality of the `port` label.

Suggest implementing `port` summarization as a by-product of the enrichment process, by which numerical port numbers are converted to human-readable labels. This process could also "collapse" multiple port numbers to a smaller set of names, thereby providing rudimentary summarization.

Building upon the `etc/services` file format, we could add support for lists and ranges (instead of single number values) to achieve this. Here is a sample file to illustrate this idea:
```
ftp              21/udp     # File Transfer [Control]
ftp              21/tcp     # File Transfer [Control]
ssh              22/udp     # SSH Remote Login Protocol
ssh              22/tcp     # SSH Remote Login Protocol
domain           53/udp     # Domain Name Server
domain           53/tcp     # Domain Name Server
...
dhcp             67,68/udp
dhcp             67,68/tcp
...
dynamic-client   49152–65535/udp
dynamic-client   49152–65535/tcp
```

leoparente:


rboucher-me (approved):

src/IpPort.h
```diff
@@ -4,13 +4,19 @@
...
 #endif
+#include <csv.h>
+#ifdef __GNUC__
+#pragma GCC diagnostic pop
+#endif
+#include <map>
 #include <nlohmann/json.hpp>
 #include <ostream>
 #include <string>
+#include <variant>
```
weyrick: where is this used?

src/IpPort.cpp
```diff
@@ -8,28 +8,67 @@ std::ostream &operator<<(std::ostream &os, const IpPort &p)
...
+                    if (protocol == "tcp") {
+                        ports_tcp_list[value] = {service, value};
+                    } else if (protocol == "udp") {
+                        ports_udp_list[value] = {service, value};
+                    }
+                }
+            }
+            break;
+        } catch ([[maybe_unused]] std::exception &e) {
+            in.next_line();
```
weyrick, line 45: should we warning here?

weyrick (approved):

src/IpPort.h
leoparente: it is not

src/IpPort.cpp
leoparente: I don't think so because it happens quite often on iana file due to multiline description for Ports like this:

```
virtualdj,,,VirtualDJ Remote Control protocol,[Atomix_Productions],[Stephane_Clavel],2011-10-12,,,,,Defined TXT keys: None
visel,,,visel Q-System services,[Lehpaner_Sasa],[Lehpaner_Sasa],,,,,,"Defined TXT keys:   Version=<service version>
  Type=<service type>
  Name=<service Name>
  Zone=<service zone>
  ID=<serviceID>
  SString=<serviceString>
  SString1=<serviceString>"
voalte2,,tcp,Server location via DNS-SD,[Voalte_Inc],[John_Simpson],2019-01-22,,,,,"Defined TXT keys: txtvers, name, (others not publicly documented)"
voalte3,,tcp,Server location via DNS-SD,[Voalte_Inc],[John_Simpson],2019-01-22,,,,,"Defined TXT keys: txtvers, name, (others not publicly documented)"
volterio,,tcp,Internal Communication Service,[Volterio],[Johannes_Ofner],2021-07-23,,,,,Defined TXT keys: None
vos,,,Virtual Object System (using VOP/TCP),[Reed_Hedges][Peter_Amstutz],[Reed_Hedges][Peter_Amstutz],,,,,,"Defined TXT keys:   url=<Object URL with full path>
  type=<Comma-seperated list of object types>
```

leoparente: So the log would be noisy

src/IpPort.h
leoparente: I removed it

PR #555, opened 2022-12-01
Feature/ip summarization

leoparente:


src/Metrics.h
```diff
@@ -457,17 +471,38 @@ class TopN final : public Metric
...
         auto section = json::array();
         auto items = _fi.get_frequent_items(datasketches::frequent_items_error_type::NO_FALSE_NEGATIVES);
         auto threshold = _get_threshold(items);
-        for (uint64_t i = 0; i < std::min(_top_count, items.size()); i++) {
-            if (items[i].get_estimate() >= threshold) {
-                section[i]["name"] = formatter(items[i].get_item());
-                section[i]["estimate"] = items[i].get_estimate();
-            } else {
-                break;
+        if (op == Aggregate::DEFAULT) {
```
weyrick, line 495: can we break this logic out so that we are not copy and pasting it between json and prometheus?

src/handlers/flow/FlowStreamHandler.h
```diff
@@ -57,7 +72,15 @@ struct DeviceEnrich {
     std::unordered_map<uint32_t, InterfaceEnrich> interfaces;
 };
 
-typedef std::unordered_map<std::string, DeviceEnrich> EnrichMap;
+struct SummaryData {
+    IpSummary type{IpSummary::None};
+    std::vector<Ipv4Subnet> ipv4_exclude_summary;
```
weyrick: Do we have a sense of the memory impact for these?

src/handlers/flow/FlowStreamHandler.cpp
```diff
@@ -14,15 +14,125 @@
 
 namespace visor::handler::flow {
 
-template <typename Out>
-static void split(const std::string &s, char delim, Out result)
+static std::pair<bool, std::vector<Ipv4Subnet>::const_iterator> match_subnet(std::vector<Ipv4Subnet> &ipv4_list, uint32_t ipv4_val)
```
weyrick: we have these matching and parsing functions in many places too by now, don't we? can we move to a common utility library for it?

weyrick (changes requested):

src/handlers/flow/FlowStreamHandler.cpp
leoparente: Yes, I think we have this in 3 places... so it would be Better to be in a lib

src/handlers/flow/FlowStreamHandler.h
leoparente: Those are vector with low data points like two or three.

And this structure is kept in MetricsManager and buckets only has a pointer reference to it

src/Metrics.h
leoparente: done

src/handlers/flow/FlowStreamHandler.cpp
leoparente: I did it

weyrick (approved):

PR #566, opened 2022-12-07
Make xact timeout configurable

### LINKED ISSUE #559: make transaction timeout user configurable
Currently TransactionManager supports a custom TTL in the constructor, but this is not exposed to end users. 

For the DNS Handler, we should expose a new handler configuration variable `xact_ttl_secs` to set this.

https://github.com/ns1labs/pktvisor/blob/db7b32d1f400273f44a222768d89e622ae5f4fd8/libs/visor_transaction/TransactionManager.h#L60

leoparente:


src/handlers/dhcp/DhcpStreamHandler.h
```diff
@@ -118,18 +118,24 @@ class DhcpMetricsBucket final : public visor::AbstractMetricsBucket
...
-    TransactionManager<uint32_t, DhcpTransaction, std::hash<uint32_t>> _request_ack_manager;
+    typedef TransactionManager<uint32_t, DhcpTransaction, std::hash<uint32_t>> DhcpTransactionManager;
+    std::unique_ptr<DhcpTransactionManager> _request_ack_manager;
 
 public:
     DhcpMetricsManager(const Configurable *window_config)
         : visor::AbstractMetricsManager<DhcpMetricsBucket>(window_config)
     {
+        if (window_config->config_exists("xact_ttl_secs")) {
+            _request_ack_manager = std::make_unique<DhcpTransactionManager>(static_cast<uint32_t>(window_config->config_get<uint64_t>("xact_ttl_secs")));
```
weyrick: what happens if it's a string, is that possible?

weyrick (approved):
looks good, just question on error handling for config item type

src/handlers/dhcp/DhcpStreamHandler.h
leoparente: This would except the as invalid value on policy creation... but it is not a good practice to except on constructor... 

What do you think of creating a do that validation on StreamHandler (ex. DhcpStreamHandler) Start? but it would generate a dependency of StreamHandler (ex. DhcpStreamHandler) creating the transaction manager on MetricsManager class (ex. DhcpMetricsManager)

PR #584, opened 2022-12-19
Spike/otel proto

### LINKED ISSUE #553: Spike - Opentelemetry client
https://conan.io/center/opentelemetry-cpp

leoparente:


cmd/pktvisord/main.cpp
```diff
@@ -100,6 +100,14 @@ static const char USAGE[] =
     Prometheus Options:
       --prometheus                          Ignored, Prometheus output always enabled (left for backwards compatibility)
       --prom-instance ID                    Optionally set the 'instance' label to given ID
+    Opentelemetry Options
```
weyrick: let's keep capital T for their branding `OpenTelemetry`

cmd/pktvisord/main.cpp
```diff
@@ -100,6 +100,14 @@ static const char USAGE[] =
...
+      --opentelemetry                       Enable Opentelemetry OTLP exporter over HTTP
```
weyrick: since everything else here is `otel` i suggest we keep this `--otel` too. also in description let's use `HTTP(S)`

cmd/pktvisord/main.cpp
```diff
@@ -100,6 +100,14 @@ static const char USAGE[] =
...
+      --otel-host HOST                      Setup OTEL Collector IP to where the data will be pushed to (default: localhost)
```
weyrick: `Set OTEL destination IP where the data will be pushed to (default: localhost)`

cmd/pktvisord/main.cpp
```diff
@@ -100,6 +100,14 @@ static const char USAGE[] =
...
+      --otel-port PORT                      Setup OTEL Collector port number (default: 4317)
```
weyrick: `Set OTEL destination port number`

cmd/pktvisord/main.cpp
```diff
@@ -100,6 +100,14 @@ static const char USAGE[] =
...
+      --otel-interval N                     The interval in seconds that exporter will periodically push data (default: 60)
+      --otel-tls                            Enable TLS communication between Exporter and Collector
```
weyrick: `Enable TLS when connecting to OTEL destination`

src/OpenTelemetry.h
```diff
@@ -0,0 +1,69 @@
...
+    uint64_t interval_sec{60};
+    std::string tls_cert;
+    std::string tls_key;
+};
+
+class OpenTelemetry
+{
+    std::unique_ptr<httplib::Client> _client;
+    collector::metrics::v1::ExportMetricsServiceRequest _request;
+    metrics::v1::ResourceMetrics *_resorce;
```
weyrick: typo, `resource`

src/OpenTelemetry.h
```diff
@@ -0,0 +1,69 @@
...
+        }
+        _resorce = _request.add_resource_metrics();
+        static timer timer_thread{std::chrono::seconds(config.interval_sec)};
+        _timer_handle = timer_thread.set_interval(std::chrono::seconds(config.interval_sec), [this] {
+            _resorce->clear_scope_metrics();
+            if (_callback && _callback(*_resorce)) {
+                if (auto body_size = _request.ByteSizeLong(); body_size > sizeof(_request)) {
+                    auto body = std::make_unique<char[]>(body_size);
+                    _request.SerializeToArray(body.get(), body_size);
+                    auto result = _client->Post("/v1/metrics", body.get(), body_size, BIN_CONTENT_TYPE);
```
weyrick: this URL doesn't need to be configurable?

src/Policies.cpp
```diff
@@ -398,6 +398,25 @@ void Policy::prometheus_metrics(std::stringstream &out)
...
+        for (auto &[bucket, hmod] : bucket_map) {
+            hmod->window_opentelemetry(scope, bucket.get(), {{"policy", name()}, {"handler", hmod->schema_key() + "_merged"}});
+        }
+    } else {
+        for (auto &mod : modules()) {
+            auto hmod = dynamic_cast<StreamHandler *>(mod);
+            if (hmod) {
+                spdlog::stopwatch sw;
+                hmod->window_opentelemetry(scope, {{"policy", name()}, {"handler", hmod->name()}});
+                spdlog::get("visor")->debug("{} window_prometheus elapsed time: {}", hmod->name(), sw);
```
weyrick: fix copy/paste `window_prometheus`

weyrick (changes requested):
this is really great work! it's a shame that so much of this is repetition and copy/paste from json/prom code (especially all the new duplicate logic involving groups in the handlers) ... we will have to save that as tech debt for now

src/OpenTelemetry.h
leoparente:
```
OTLP/HTTP Request
Telemetry data is sent via HTTP POST request. The body of the POST request is a payload either in binary-encoded Protobuf format or in JSON-encoded Protobuf format.

The default URL path for requests that carry trace data is /v1/traces (for example the full URL when connecting to "example.com" server will be https://example.com/v1/traces). The request body is a Protobuf-encoded ExportTraceServiceRequest message.

The default URL path for requests that carry metric data is /v1/metrics and the request body is a Protobuf-encoded ExportMetricsServiceRequest message.

The default URL path for requests that carry log data is /v1/logs and the request body is a Protobuf-encoded ExportLogsServiceRequest message.

The client MAY gzip the content and in that case MUST include "Content-Encoding: gzip" request header. The client MAY include "Accept-Encoding: gzip" request header if it can receive gzip-encoded responses.

Non-default URL paths for requests MAY be configured on the client and server sides.
```
yeah, I think it should be configurable as well

cmd/pktvisord/main.cpp
leoparente: done

src/OpenTelemetry.h
leoparente: I made it configurable

src/OpenTelemetry.h
leoparente: done

src/Policies.cpp
leoparente: done

cmd/pktvisord/main.cpp
leoparente: done

leoparente:
> this is really great work! it's a shame that so much of this is repetition and copy/paste from json/prom code (especially all the new duplicate logic involving groups in the handlers) ... we will have to save that as tech debt for now

Agree. I think we could find a way to use template to handle `to_json/prom/otel `functions

weyrick (approved):

PR #587, opened 2022-12-21
Netprobe: execute the _recv callback process on correct thread

### LINKED ISSUE #554: Response time for ping test on netprobe quantiles metrics increases as time goes by
When an agent with netprobe tap is running a netprobe policy, the response time for "ping" test increases as time goes by and it is weird.

I have two agents running and both have similar behavior:

I have another agent, on the same machine, that starts after the 2 mentioned.. the pattern looks similar, but the response time is different:

image

`/api/v1/policies/policy-probe/metrics/prometheus` data:

[agent1_policy-probe.txt](https://app.zenhub.com/files/241899022/acb9f8a6-8bad-42c3-813c-1994ace116c3/download)

[agent0_policy-probe.txt](https://app.zenhub.com/files/241899022/d0269083-73fb-4f5f-8802-a126746fdf17/download)

`/api/v1/policies/__all/metrics/prometheus` data:

[agent1_all.txt](https://app.zenhub.com/files/241899022/2a15662e-f29c-475d-95fe-a82d8b76f7ed/download)

[agent0_all.txt](https://app.zenhub.com/files/241899022/acaf82b6-36bf-4e02-baad-3c2fd9fba234/download)

To reproduce:

Agent tap:

```
  taps:
    default_netprobe:
      input_type: netprobe
      config:
        test_type: ping
      tags:
        virtual: true
        vhost: 1
```

Policy:

```
handlers:
  modules:
    default_netprobe:
      type: netprobe
      metric_groups:
        enable:
          - counters
          - quantiles
input:
  tap: default_netprobe
  input_type: netprobe
  config:
    test_type: ping
    packets_per_test: 5
    interval_msec: 2000
    timeout_msec: 1000
    packets_interval_msec: 10
    packet_payload_size: 56
    targets:
      www.google.com:
        target: www.google.com
      'orb live':
        target: orb.live
      orb_community:
        target: orb.community       
kind: collection
```

grafana query: `avg(netprobe_response_quantiles_us{quantile="0.5"}) by (agent,target, policy)`

obs: i'm usign the query filtering the 'orb live' data because it have another problem so..(`avg(netprobe_response_quantiles_us{quantile="0.5", target!="orb live"}) by (agent,target, policy)`)

leoparente:
- _recv callback for every target was being executed by the static Receiver thread
    - Moved the cost of execution to the PingProbe thread

### DIFF
--- a/src/inputs/netprobe/PingProbe.cpp
+++ b/src/inputs/netprobe/PingProbe.cpp
@@ -1,6 +1,7 @@
 #include "PingProbe.h"
 
 #include "NetProbeException.h"
+#include "ThreadName.h"
 #include <Packet.h>
 #include <TimespecTimeval.h>
 #include <uvw/idle.h>
@@ -8,7 +9,6 @@
 namespace visor::input::netprobe {
 
 sigslot::signal<pcpp::Packet &, timespec> PingReceiver::recv_signal;
-std::recursive_mutex PingReceiver::mutex;
 thread_local std::atomic<uint32_t> PingProbe::sock_count{0};
 thread_local SOCKET PingProbe::_sock{INVALID_SOCKET};
 
@@ -94,7 +94,6 @@ void PingReceiver::_setup_receiver()
                 TIMESPEC_TO_TIMEVAL(&time, &stamp);
                 pcpp::RawPacket raw(reinterpret_cast<uint8_t *>(_array.data()), rc, time, false, pcpp::LINKTYPE_DLT_RAW1);
                 pcpp::Packet packet(&raw, pcpp::ICMP);
-                std::lock_guard<std::recursive_mutex> lock(mutex);
                 recv_signal(packet, stamp);
             }
         }
@@ -105,6 +104,7 @@ void PingReceiver::_setup_receiver()
 
     // spawn the loop
     _io_thread = std::make_unique<std::thread>([this] {
+        thread::change_self_name("receiver", "ping");
         _io_loop->run();
     });
 }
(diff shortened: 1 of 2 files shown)


leoparente:
Before this PR (current develop)
Static Receiver thread is executing the Packet receiver process end to end: from socket receive until Handler metric generation:
image
As we can see, the thread that spends more CPU time is the receiver thread
This test has two Netprobe policies with ping test

leoparente:
The change that was done was to add a receiver vector owned by each PingProbe Instance that is filled by the receiver thread.
So the _recv process (Metrics generated on Handler) is now done by the thread that the  PingProbe Instance is running on:
Screenshot from 2022-12-21 18-10-06

leoparente:
I will work on a optimization to have a single static vector

src/inputs/netprobe/PingProbe.cpp
```diff
@@ -93,18 +93,29 @@ void PingReceiver::_setup_receiver()
                 timeval time;
                 TIMESPEC_TO_TIMEVAL(&time, &stamp);
                 pcpp::RawPacket raw(reinterpret_cast<uint8_t *>(_array.data()), rc, time, false, pcpp::LINKTYPE_DLT_RAW1);
-                pcpp::Packet packet(&raw, pcpp::ICMP);
-                std::lock_guard<std::recursive_mutex> lock(mutex);
-                recv_signal(packet, stamp);
+                _recv_packets.push_back({pcpp::Packet(&raw, pcpp::ICMP), stamp});
```
weyrick: do you look at the Packet raw data ownership model here? I recall there being various options to have the raw data freed or not when this Packet object is destroyed - just want to make sure we don't either double free or leak

src/inputs/netprobe/PingProbe.cpp
```diff
@@ -173,17 +186,25 @@ bool PingProbe::start(std::shared_ptr<uvw::Loop> io_loop)
...
+    _recv_handler = _io_loop->resource<uvw::AsyncHandle>();
+    if (!_recv_handler) {
+        throw NetProbeException("PingProbe - unable to initialize AsyncHandle receiver");
+    }
+    _recv_handler->on<uvw::AsyncEvent>([this](const auto &, auto &) {
+        for (auto &[packet, stamp] : PingReceiver::recv_packets) {
+            _recv(packet, TestType::Ping, _name, stamp);
+        }
+    });
+    _receiver->register_async_callback(_recv_handler);
```
weyrick, line 205: so how often does/when this fire?

leoparente: I kept the 100ms timer on receiver thread... so it will be called every 100ms. But now on the right thread

src/inputs/netprobe/PingProbe.cpp
leoparente: yes.. I did try other property and got double freed issue

leoparente: I've also ran the valgrind to check for memory leak and got none

weyrick (approved):

PR #588, opened 2022-12-22
Implement exact match Qname filter and make improvements on O(1) filt…

### LINKED ISSUE #582: Exact match QNAME filter
As a DNS administrator,
I would like to be able to collect and view metrics for a specific QNAME,
So that I can get detailed analysis and insights related exclusively to that QNAME.

Currently we support filtering based on `only_qname_suffix`, which isn't limited to a specific QNAME and includes all QNAMES that end with the specified suffixes.

The target solution would allow filtering against a list of specific QNAMES and require an exact match:
```
filter:
  only_qnames: [ www.foo.bar, blah.example.org]
```

leoparente:
…ering

### DIFF
--- a/src/handlers/dns/v2/DnsStreamHandler.cpp
+++ b/src/handlers/dns/v2/DnsStreamHandler.cpp
@@ -158,15 +158,6 @@ void DnsStreamHandler::start()
             _f_qnames.emplace_back(std::move(qname_ci));
         }
     }
-    if (config_exists("only_qname")) {
-        _f_enabled.set(Filters::OnlyQName);
-        for (const auto &qname : config_get<StringList>("only_qname")) {
-            std::string qname_ci{qname};
-            std::transform(qname_ci.begin(), qname_ci.end(), qname_ci.begin(),
-                [](unsigned char c) { return std::tolower(c); });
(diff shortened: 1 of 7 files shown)


src/handlers/dns/v2/DnsStreamHandler.cpp
```diff
@@ -149,16 +149,31 @@ void DnsStreamHandler::start()
...
+    if (config_exists("only_qname")) {
+        _f_enabled.set(Filters::OnlyQName);
+        for (const auto &qname : config_get<StringList>("only_qname")) {
+            std::string qname_ci{qname};
+            std::transform(qname_ci.begin(), qname_ci.end(), qname_ci.begin(),
+                [](unsigned char c) { return std::tolower(c); });
+            _f_qnames.emplace_back(std::move(qname_ci));
+        }
+    }
+    if (config_exists("only_qname")) {
```
weyrick: this block is here twice?

weyrick (approved):
check that one block, otherwise lgtm

src/handlers/dns/v2/DnsStreamHandler.cpp
leoparente: it was. I removed it

PR #599, opened 2023-01-08
[PKT-61] Bugfix/netflow support ipv6

leoparente:


weyrick:
I see this involves some pretty low level bit copying. We'll need to make sure we have enough test cases to cover proper bounds checking, ensuring things aren't null, have enough space to copy to, etc. Let's make sure invalid/unexpected input cannot cause a failure here.

leoparente:
> I see this involves some pretty low level bit copying. We'll need to make sure we have enough test cases to cover proper bounds checking, ensuring things aren't null, have enough space to copy to, etc. Let's make sure invalid/unexpected input cannot cause a failure here.

Yes, Agreed. That is why I used std::array with enough space for IPV4 and IPV6.. and regarding Netflow V9 and IPFIX `be_copy` logic ensures that: only copy if there is enough space

weyrick (approved):

PR #604, opened 2023-01-10
Implement OTLP Histogram

leoparente:
Implemented based on Prometheus and guides provided by OpenTelemetry

> Histograms
A [Prometheus Histogram](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram) MUST be converted to an OTLP Histogram. Multiple Prometheus histogram metrics MUST be merged together into a single OTLP Histogram:
1, The le label on non-suffixed metrics is used to identify and order histogram bucket boundaries. Each Prometheus line produces one bucket count on the resulting histogram. Each value for the le label except +Inf produces one bucket boundary.
2, Lines with _count and _sum suffixes are used to determine the histogram’s count and sum.
3, If _count is not present, the metric MUST be dropped.
4, If _sum is not present, the histogram’s sum MUST be unset.

### DIFF
--- a/src/Metrics.h
+++ b/src/Metrics.h
@@ -296,13 +296,33 @@ public:
 
     void to_opentelemetry(metrics::v1::ScopeMetrics &scope, timespec &start, timespec &end, LabelMap add_labels = {}) const
     {
+        if (_sketch.is_empty()) {
+            return;
+        }
+        auto bins_pmf = _get_boundaries();
+        auto histogram_pmf = _sketch.get_PMF(bins_pmf.first.data(), bins_pmf.second);
+        std::vector<T> bins;
+        for (size_t i = 0; i < bins_pmf.second; ++i) {
+            if (histogram_pmf[i]) {
+                bins.push_back(bins_pmf.first[i]);
+            }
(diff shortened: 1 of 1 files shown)


leoparente:
Tested with exporter
```
Metric #5
Descriptor:
     -> Name: netprobe_response_histogram_us
     -> Description: Net Probe histogram in microseconds
     -> Unit: 
     -> DataType: Histogram
     -> AggregationTemporality: Unspecified
HistogramDataPoints #0
Data point attributes:
     -> handler: Str(default2-default_netprobe-net)
     -> policy: Str(default2)
     -> tap: Str(default_netprobe)
     -> target: Str(outlook)
StartTimestamp: 2023-01-10 20:42:12.399318495 +0000 UTC
Timestamp: 1970-01-01 00:00:00 +0000 UTC
Count: 1923
ExplicitBounds #0: 100000.000000
ExplicitBounds #1: 113646.000000
Buckets #0, Count: 1339
Buckets #1, Count: 1923
```

weyrick (approved):

PR #609, opened 2023-01-13
Implement chaining flow for TCP packets

leoparente:


src/handlers/net/NetStreamHandler.cpp
```diff
@@ -157,6 +194,22 @@ void NetStreamHandler::set_end_tstamp(timespec stamp)
     }
 }
 
+bool NetStreamHandler::validate_tcp_data(const pcpp::ConnectionData &connectionData, PacketDirection dir, timeval timeInterval)
+{
+    pcpp::Packet packet;
+    if (connectionData.srcIP.isIPv4()) {
+        packet.addLayer(new pcpp::IPv4Layer(connectionData.srcIP.getIPv4(), connectionData.dstIP.getIPv4()), true);
```
weyrick, line 201: are we sure this memory is freed correctly?

weyrick (approved):
Can you summarize the strategy implemented here in the PR description before landing please?

src/handlers/net/NetStreamHandler.cpp
leoparente: yes. This transfer ownership to `packet` object

```
		/**
		 * Add a new layer as the last layer in the packet. This method gets a pointer to the new layer as a parameter
		 * and attaches it to the packet. Notice after calling this method the input layer is attached to the packet so
		 * every change you make in it affect the packet; Also it cannot be attached to other packets
		 * @param[in] newLayer A pointer to the new layer to be added to the packet
		 * @param[in] ownInPacket If true, Packet fully owns newLayer, including memory deletion upon destruct.  Default is false.
		 * @return True if everything went well or false otherwise (an appropriate error log message will be printed in
		 * such cases)
		 */
		bool addLayer(Layer* newLayer, bool ownInPacket = false) { return insertLayer(m_LastLayer, newLayer, ownInPacket); 
		/**
```

leoparente: If you pass a object instead a pointer and do ownInPacket = true, double free happens

leoparente:
> Can you summarize the strategy implemented here in the PR description before landing please?

@weyrick I will do a explanation thread here

Firstly, let start with Pktvisor PCAP flow:

Team document

leoparente:
Here is the sample and explanation. after doing the explanation... I thought more about the implementation and I think we should rename the new `tcp_signal` to `tcp_reassembled_signal`

Team document (1)

PR #613, opened 2023-01-17
Add support to subnet wildcard summary

### LINKED ISSUE #583: IP summarization based on subnet CIDR
As a network traffic engineer,
I would like to be able to easily and succinctly configure how to summarize flow traffic by IP using subnet CIDRs,
So that I can more easily observe network traffic trends, reduce the volume of data I need to collect and improve the responsiveness of my observability tools.

This feature would allow defining IP summarization by subnets using broad rules without having to explicitly identify all subnets that should be summarized.

The target solution would support configuring IP summarization with a configuration such as:
```
summarize_ips_by_cidr: true
cidr_summarization:
        - 1-10.0.0.0/8
        - 172.16-31.0.0/16 
	- 0.0.0.0/24      # wildcard meaning all other IPs should be summarized as /24 (e.g. "a.b.c.0")
exclude_ips_from_summarization:
	- 10.1.0.0/16     # exclude all IPs in this subnet from summarization
	- 192.168.1-255.0/24
```

leoparente:


src/handlers/flow/FlowStreamHandler.cpp
```diff
@@ -40,17 +41,37 @@ static std::string ip_summarization(const std::string &val, SummaryData *summary
...
+            if (!check_subnet) {
+                return asn;
+            }
+        }
+        if (summary->type == IpSummary::BySubnet || check_subnet) {
             if (ipv4.isValid()) {
                 if (auto subnet = match_subnet(summary->ipv4_summary, ipv4.toInt()); subnet.has_value()) {
                     return subnet.value()->str;
+                } else if (summary->ipv4_wildcard.has_value()) {
+                    return pcpp::IPv4Address(ipv4.toInt() & (htobe32((0xFFFFFFFFu) << (32 - summary->ipv4_wildcard.value().cidr)))).toString()
```
weyrick: I would prefer to have this in a separate utility function - seems like something we should have in a common place.

src/handlers/flow/FlowStreamHandler.cpp
```diff
@@ -40,17 +41,37 @@ static std::string ip_summarization(const std::string &val, SummaryData *summary
...
+                        + "/" + std::to_string(summary->ipv4_wildcard.value().cidr);
                 }
             } else if (ipv6.isValid()) {
                 if (auto subnet = match_subnet(summary->ipv6_summary, ipv6.toBytes()); subnet.has_value()) {
                     return subnet.value()->str;
+                } else if (summary->ipv6_wildcard.has_value()) {
+                    uint8_t bits = 128 - summary->ipv6_wildcard.value().cidr;
```
weyrick: same here

weyrick (changes requested):

src/handlers/flow/FlowStreamHandler.cpp
leoparente: done

src/handlers/flow/FlowStreamHandler.cpp
leoparente: done

weyrick (approved):

PR #615, opened 2023-01-18
Add support to Net Handler V2

leoparente:


src/handlers/net/v2/NetStreamHandler.h
```diff
@@ -0,0 +1,252 @@
...
+    Rate _rate_total;
+    Rate _throughput_in;
+    Rate _throughput_out;
+    Rate _throughput_total;
+
+    void _process_geo_metrics(const pcpp::IPv4Address &ipv4);
+    void _process_geo_metrics(const pcpp::IPv6Address &ipv6);
+
+public:
+    NetworkMetricsBucket()
```
leoparente, line 176: @rboucher-me I would like your help here to name the metrics

I've alredy fix the geo_loc, asn and removed "payload" metrics name

@weyrick the idea here is to add a new Net Handler with proper fixes on metric names

src/handlers/net/v2/NetStreamHandler.h
```diff
@@ -0,0 +1,252 @@
...
+        Counter total_unk;
+        Counter total;
+        Counter filtered;
+        counters()
+            : UDP(NET_SCHEMA, {"udp_packets"}, "Count of UDP packets")
+            , TCP(NET_SCHEMA, {"tcp_packets"}, "Count of TCP packets")
+            , OtherL4(NET_SCHEMA, {"other_l4_packets"}, "Count of packets which are not UDP or TCP")
+            , IPv4(NET_SCHEMA, {"ipv4_packets"}, "Count of IPv4 packets")
+            , IPv6(NET_SCHEMA, {"ipv6_packets"}, "Count of IPv6 packets")
+            , TCP_SYN(NET_SCHEMA, {"protocol", "tcp", "syn_packets"}, "Count of TCP SYN packets")
```
leoparente: how should we proper rename this metric? @rboucher-me  the current prometheus name became
`net_protocol_tcp_syn_packets`

rboucher-me: Ideally we would just name it `net_tcp_syn_packets`. Does the `protocol` have to be there?

src/handlers/net/v2/NetStreamHandler.h
```diff
@@ -0,0 +1,252 @@
...
+    Rate _throughput_in;
+    Rate _throughput_out;
+    Rate _throughput_total;
+
+    void _process_geo_metrics(const pcpp::IPv4Address &ipv4);
+    void _process_geo_metrics(const pcpp::IPv6Address &ipv6);
+
+public:
+    NetworkMetricsBucket()
+        : _srcIPCard(NET_SCHEMA, {"cardinality", "src_ips_in"}, "Source IP cardinality")
```
rboucher-me: wondering what the `_in` means here? are we really looking at just inbound traffic? if not, I think we should drop the `_in`

src/handlers/net/v2/NetStreamHandler.h
```diff
@@ -0,0 +1,252 @@
...
+        , _dstIPCard(NET_SCHEMA, {"cardinality", "dst_ips_out"}, "Destination IP cardinality")
```
rboucher-me: same as above, but for `_out`

src/handlers/net/v2/NetStreamHandler.h
```diff
@@ -0,0 +1,252 @@
...
+        , _topGeoLoc(NET_SCHEMA, "geo_loc", {"top_geo_loc_packets"}, "Top GeoIP locations")
+        , _topASN(NET_SCHEMA, "asn", {"top_asn_packets"}, "Top ASNs by IP")
+        , _topIPv4(NET_SCHEMA, "ipv4", {"top_ipv4_packets"}, "Top IPv4 IP addresses")
```
rboucher-me: don't need the "IP" in "Top IPv4 IP addresses" ("Top IPv4 addresses")

src/handlers/net/v2/NetStreamHandler.h
```diff
@@ -0,0 +1,252 @@
...
+        , _topIPv6(NET_SCHEMA, "ipv6", {"top_ipv6_packets"}, "Top IPv6 IP addresses")
```
rboucher-me: same as above

src/handlers/net/v2/NetStreamHandler.h
```diff
@@ -0,0 +1,252 @@
...
+        , _payload_size(NET_SCHEMA, {"payload_size_bytes"}, "Quantiles of payload sizes, in bytes")
+        , _rate_in(NET_SCHEMA, {"rates", "in_pps"}, "Rate of ingress in packets per second")
```
rboucher-me: should be "Rate of ingress packets per second"

src/handlers/net/v2/NetStreamHandler.h
```diff
@@ -0,0 +1,252 @@
...
+        , _rate_out(NET_SCHEMA, {"rates", "out_pps"}, "Rate of egress in packets per second")
```
rboucher-me: should be "Rate of egress packets per second"

weyrick (approved):

PR #616, opened 2023-01-19
Perform summary for each flow packet and add LRUCache for summary data

leoparente:


leoparente:
Flow Handler unit tests are running faster with this approach

weyrick (approved):
lgtm, are there any concerns about threads here?

leoparente:
> lgtm, are there any concerns about threads here?

No. it is all separated by thread context, so no data exchanged between threads

PR #637, opened 2023-02-24
Remove some unused + Minor if/else fix

egecetin:


leoparente (approved):
@egecetin thanks for the contrib! Let me know if you have any doubts about the project.

PR #640, opened 2023-03-09
specify required Conan version in README

emn:
While building pktvisor I encountered an error which led me to discover that pktvisor will not build with the latest version of Conan. In Conan 2.X, `--install-folder` has been [deprecated](https://docs.conan.io/1/migrating_to_2.0/commands.html). I downgraded my local Conan to version 1.59.0 (the latest pre-2.X release) and was able to build successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -591,7 +591,7 @@ pktvisor is developed and tested on Linux and OSX. A Windows port is in progress
 
 #### Dependencies
 
-* [Conan](https://conan.io/) C++ package manager
+* [Conan](https://conan.io/) 1.X C++ package manager
 * CMake >= 3.13 (`cmake`)
 * C++ compiler supporting C++17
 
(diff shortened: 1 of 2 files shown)


emn:
Conan 2.0.1 is now the version being distributed by pip, which means that your Docker builds are also broken. It would be simple enough to specify a 1.X version in the Dockerfile, but I assume you will want to fix this properly. There is a [migration guide](https://docs.conan.io/1/conan_v2.html) but unfortunately I know nothing about Conan (or CMAKE for that matter) so I can't contribute the fix myself.

leoparente:
Yes, you are right @emn we have fixed our pipeline to work with conan 1.59 for now but we forget to update all the docs. Thanks for your contrib. I've opened a issue to do this migration. https://github.com/orb-community/pktvisor/issues/641

leoparente (approved):

PR #650, opened 2023-03-23
rcode = response code

manrodrigues:
Terry suggested an update to the definition of RCODE (https://github.com/orb-community/orb-website/pull/45#issue-1609346314) to match with rfc (https://www.rfc-editor.org/rfc/rfc1035.html).


leoparente (approved):
LGTM

PR #662, opened 2023-04-13
Feature/limit tcp reassembly

leoparente:


RFCs/2021-04-16-76-collection-policies.md
```diff
@@ -38,6 +38,8 @@ visor:
         input_type: pcap
         filter:
           bpf: "port 53"
+        config:
+          tcp_reassembly_limit: 100000
```
leoparente: @weyrick @rboucher-me i made the tcp reassembly packets size configurable.. I don't know the best name for this config... what do you guys think?

etaques: seems good this name to me :smiley:

weyrick: if it's essentially the number of concurrent TCP connections we support, that I suggest we instead make it "tcp_stream_limit", since "reassembly" isn't very intuitive for users

leoparente: @weyrick I agree with change but it is actually the number of tcp packages... for instance, a single tcp connection can have more than 3 packages in this structure (syn, ack, actual data..) the 'real' name would be:
`tcp_packets_reassembly_limit`. what would be a intuitive name for end users?

rboucher-me: Though a bit long, `tcp_packet_reassembly_cache_limit` feels more intuitive to me (`cache` could be replaced with something else such as `buffer` or ?)

weyrick: fine with that. almost no one will need to change

leoparente: done. Went with `tcp_packet_reassembly_cache_limit`

etaques (approved):
LGTM

PR #667, opened 2023-04-19
[Partial Fix] - remove/comment end timestamp from OTLP export metrics

leoparente:
This is not the full solution yet. But we need to test it end to end with orb agent.


etaques (approved):
LGTM

PR #678, opened 2023-06-01
Ensure TopN _item_key label has proper value before adding opentelemetry datapoint

leoparente:


etaques (approved):
Many thanks for it! I will check whether the problem is not on private module before try this solution

PR #684, opened 2023-06-25
Upgrade datasketches lib from 3.3.0 to 4.1.0

leoparente:
Quantile inclusive weight by default now. That means no required pace workaround anymore.
Also Fixed float implicit conversions and some compiler warnings.


leoparente:
Doing other project, I've noticed that sometimes implicit conversions are not optimized by the compiler and can impact a bit in performance, even more if it is in a critical path.

weyrick (approved):

PR #689, opened 2023-09-22
migrate from catch2v2 to catch2v3 and implement VisorTest helper lib

leoparente:
### Migrate pktvisor tests from Catch2 v2 to Catch2 v3
 - Perform the upgrade based on: https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md.
    - PS: with that change, no main.cpp is required anymore for tests
 - Created `LibVisorTest` to initialize spdlog whenever it is required using Catch2 v3 standard.
 - Reduced folders and files complexity by migrating test cmake commands to Module CmakeFile


weyrick (approved):
great, thanks for this!

PR #699, opened 2023-11-01
OTLP: Move policy name to attributes

leoparente:
- Move pktvisor policy_name to attributes.
- No name is set for the scope anymore (i'm not sure if it is required)

- It requires validation! cc @manrodrigues

### DIFF
--- a/src/CoreServer.cpp
+++ b/src/CoreServer.cpp
@@ -448,6 +448,7 @@ void CoreServer::_setup_routes(const PrometheusConfig &prom_config)
                 try {
                     auto [policy, lock] = _registry->policy_manager()->module_get_locked(p_mname);
                     auto scope = resource.add_scope_metrics();
+                    scope->mutable_scope()->set_name("pktvisor/" + p_mname);
                     auto attr = scope->mutable_scope()->add_attributes();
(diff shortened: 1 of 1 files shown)


src/CoreServer.cpp
```diff
@@ -448,7 +448,9 @@ void CoreServer::_setup_routes(const PrometheusConfig &prom_config)
                 try {
                     auto [policy, lock] = _registry->policy_manager()->module_get_locked(p_mname);
                     auto scope = resource.add_scope_metrics();
-                    scope->mutable_scope()->set_name(p_mname);
+                    auto attr = scope->mutable_scope()->add_attributes();
```
lpegoraro, line 452: nice!
with this the scope name should have something like "pktvisorreceiver"
right @weyrick ?

weyrick: @leoparente can we make `scope->mutable_scope()->set_name();` "pktvisor/POLICYNAME" please?

leoparente: sure

leoparente: done

weyrick (approved):

PR #717, opened 2024-08-01
chore: update pktvisor conan dependencies

leoparente:
- Disable Armv7 build until conan v2 upgrade
- Disable integration tests until orb.live is online again


weyrick:
are these test fails unexpected?

leoparente:
@weyrick .No, it is not expected and current build is broken. I'm fixing it first, so we can update dependencies.

weyrick (approved):

PR #727, opened 2024-12-03
chore: remove conan folder before running codeQL analysis

leoparente:


.github/workflows/code-ql.yml
```diff
@@ -104,6 +107,8 @@ jobs:
         cmake -DCMAKE_BUILD_TYPE=Release ..
         # build and run tests
         make all test
+        # remove conan cache
```
weyrick: do we not want this instead or also? https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#avoiding-unnecessary-scans-of-pull-requests

leoparente: > on:pull_request:paths-ignore and on:pull_request:paths set conditions that determine whether the actions in the workflow will run on a pull request. They don't determine what files will be analyzed when the actions are run. When a pull request contains any files that are not matched by on:pull_request:paths-ignore or on:pull_request:paths, the workflow runs the actions and scans all of the files changed in the pull request, including those matched by on:pull_request:paths-ignore or on:pull_request:paths, unless the files have been excluded. For information on how to exclude files from analysis, see "[Specifying directories to scan](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#specifying-directories-to-scan)."

This is related only to files changes that would trigger or not the code analysis job, not the files that will be scanned or not

leoparente: https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#specifying-directories-to-scan

oh we can do this one> I will try it as well

.github/workflows/code-ql.yml
```diff
@@ -2,6 +2,9 @@
 name: "CodeQL"
 
 on:
+  pull_request:
```
leoparente: this is only for testing purpose. It will be removed

.github/workflows/code-ql.yml
leoparente: This need /\ to be specified in the config file and we are currently using 

```yaml
     # queries: security-extended,security-and-quality
        queries: ns1/NS1QLPacks/codeql/cpp/NS1-security.qls@main
        config-file: ns1/NS1QLPacks/codeql/NS1-codeql-config.yml@main
```

weyrick: oh! we need to change this to our own config then

leoparente: yes testing it

leoparente: > Note: C/C++ does not support path-based filtering. The "paths" and "paths-ignore" configuration properties will have no effect for this language. If desired, you can use the advanced-security/filter-sarif Action to rewrite the SARIF file to exclude alerts from these paths. For more information, see https://github.com/advanced-security/filter-sarif

leoparente: lets go with the first approach of removing conan_home. I've also added our own config

weyrick (approved):
👍

PR #730, opened 2025-01-02
chore: codeql sarif filter

leoparente:
remove build/ folder from sarif scan

### DIFF
--- a/.github/workflows/code-ql.yml
+++ b/.github/workflows/code-ql.yml
@@ -2,6 +2,9 @@
 name: "CodeQL"
 
 on:
+  pull_request:
+    branches:
+      - develop
   schedule:
   #run 5 minutes after midnight daily
     - cron: '5 0 * * *'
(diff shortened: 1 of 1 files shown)


.github/workflows/code-ql.yml
```diff
@@ -108,3 +108,27 @@ jobs:
 
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v3
+      with:
+        category: "/language:cpp"
+        output: sarif-results
+        upload: failure-only
+
+    - name: filter-sarif
+      uses: advanced-security/filter-sarif@v1
```
mfiedorowicz, line 117: I think using tags for non-immutable github actions might raise "Unpinned tag for a non-immutable Action in workflow"

.github/workflows/code-ql.yml
```diff
@@ -108,3 +108,27 @@ jobs:
...
+      with:
+        patterns: |
+          -pktvisor/build/conan_home/**/*
+        input: sarif-results/cpp.sarif
+        output: sarif-results/cpp.sarif
+
+    - name: Upload SARIF
+      uses: github/codeql-action/upload-sarif@v3
```
mfiedorowicz, line 125: Is it immutable action?

leoparente: I think both are as they are maintained by github and not third party

mfiedorowicz: OK, let's try then!

mfiedorowicz (approved):

PR #769, opened 2026-04-11
chore: upgrade corrade to cci.20260327 to fix macOS build

leoparente:
## Summary

- Pins corrade to commit `22e7ffc6` (2026-03-27) via a local conan recipe at `conan/corrade/`, replacing `2020.06` which fails to compile on modern macOS
- Updates PluginManager API usage across all 19 plugin headers: constructor params `const std::string&` → `const Corrade::Containers::StringView`, `pluginInterface()` return type, and `pluginSearchPaths()` return type
- Updates `CoreRegistry.cpp` to handle `StringView`/`StringIterable` return types from `pluginList()` and `provides()`
- Removes the `-include vector` macOS workaround from CI; adds a `conan export` step for the local recipe to all four build workflows

## Test plan

- [x] macOS CI (`unit-tests-mac`) builds and passes
- [x] Linux CI (`unit-tests-linux`) still builds and passes
- [x] Windows CI (`build-win64`) still builds
- [x] Cross-compilation CI (`build_cross`) still works


leoparente:
@codex review

leoparente (commented):
All review comments addressed in commit 18b8bc5:

**Copilot: Export pinned Corrade recipe in every Conan entrypoint** — Fixed. Added `conan export` step to `build-win64` (build-release.yml) and `build-app-image-x64` (build-develop.yml). Fixed `.github/actions/build-cpp/entrypoint.sh` to copy `conan/` directory and run `conan export` before cmake (covers all Docker-based arm64/debug jobs).

**Copilot: `find_program` malformed in conan-corrade-vars.cmake** — Partially addressed. The `PATHS ENV PATH` syntax is actually valid CMake (ENV followed by variable name is a path modifier), so no change needed there. Added the recommended `NOTFOUND` guard before `get_filename_component`/`add_executable` to fail early with a clear message when cross-compiling without corrade-rc on PATH.

**Copilot: `required_conan_version = ">=1.52.0"` should be `>=2.0`** — Fixed.

**Copilot: Cache keys should include `conan/corrade/**`** — Fixed in all workflow files (build-develop.yml, build-release.yml, build_debug.yml, build_cross.yml).

**Codex P1: build_cross.yml conan export path wrong** — Fixed (`src/conan/corrade/` → `conan/corrade/`). This was causing the actual CI failure.

**Additional fix (CI failure):** `unit-tests-mac` was failing with `initializer_list not found` because Conan's CMakeToolchain sets `CMAKE_OSX_SYSROOT=macosx` (bare string) when building corrade from source. Fixed by appending `tools.apple:sdk_path` to the conan profile after `conan profile detect`, so the toolchain uses the resolved SDK path.

leoparente:
@codex review

leoparente:
Fixed in commit 0bf6279: the checkout uses `path: src` so the recipe lives at `${{github.workspace}}/src/conan/corrade/`. Updated both the export path and the cache key `hashFiles` pattern to use `src/conan/corrade/**`.

leoparente:
@codex review

leoparente:
Fixed in commit d51f546: added `conan export` to all 4 remaining paths:
- `build-release.yml` `prebuild-package` — after `get-conan`, before cmake configure
- `build-release.yml` `release` — after `get-conan`, before cmake configure
- `build_cross.yml` `pkvisor-cli` — added `pip install conan` + export before cmake (conan wasn't installed in that job at all)
- `code-ql.yml` — export runs inside the cloned `pktvisor/` directory, after `git clone`, before `cmake`

leoparente:
@codex review

leoparente:
Root cause found for cross-compilation failure: `.conanrc` sets `conan_home=./build` (relative path). The `pkvisor` job checks out into `src/` and runs `conan install` with `working-directory: src`, so its conan home resolves to `src/build/`. The export step had no `working-directory`, so it ran from the workspace root and wrote the recipe to `build/` — a completely different directory. `conan install` then correctly reported the recipe missing.

Fixed in fc16118: added `working-directory: ${{github.workspace}}/src` to the export step so both steps share the same conan home.

leoparente:
@codex review

leoparente:
Fixed cross-compile error in commit 1b7698c.

**Root cause:** `ErrorString.cpp` in Corrade selects between two `strerror_r` variants using a preprocessor condition. The POSIX branch (correct for musl) fires when `!_GNU_SOURCE`. But GCC defines `_GNU_SOURCE` by default on Linux even when the target libc is musl — so the condition is false and it falls into the GNU branch (`const char* = strerror_r(...)`). musl only implements the XSI variant (returns `int`), causing the compile error.

**Fix:** Added a patch to the local Corrade recipe (`conan/corrade/patches/0001-fix-strerror_r-musl-compat.patch`) that extends the POSIX branch condition with `|| (defined(__linux__) && !defined(__GLIBC__))`, covering musl and any other non-glibc Linux libc.

leoparente:
@codex review

leoparente:
Fixed in commit 328b63c:

**Broken patch (strerror_r musl fix):** Replaced the patch file with `replace_in_file()` in `conanfile.py`. The patch failed because patch-ng couldn't match the context lines (offset was wrong). `replace_in_file` does an exact string search regardless of line numbers — more robust for vendored recipes.

**Missing `StringStl.h` in AbstractPlugin.cpp and Policies.cpp:** Added `#include <Corrade/Containers/StringStl.h>` to both files. The `std::string(plugin())` calls in these TUs relied on the include being pulled in transitively; now it's explicit. Build verified locally (100% success).

leoparente:
@codex review

leoparente:
**P1 (Taps.h JSON serialization)** — Fixed in commit 2670aee: added `#include <Corrade/Containers/StringStl.h>` to `Taps.h`. Both `plugin()` and `pluginInterface()` return `StringView` and are assigned directly to JSON values in `info_json()`. The include was being pulled in transitively by `CORRADE_BUILD_DEPRECATED` on some compilers, now it's explicit.

**P2 (local builds require manual conan export)** — Won't fix / by design. The vendored recipe is intentional: `corrade/cci.20260327` is not published to conan-center, so a one-time `conan export conan/corrade/ --name=corrade --version=cci.20260327` is required before the first local build. This is equivalent to what CI does. The PR description and README build instructions cover the local build steps.

leoparente:
@codex review

leoparente:
Fixed in commit 93710a0:
- `docker/Dockerfile`: added `COPY ./conan/ /pktvisor-src/conan/` and `conan export conan/corrade/ --name=corrade --version=cci.20260327` before cmake
- `README.md`: added the export step to the local build instructions with a comment explaining the recipe is vendored and not in conan-center

leoparente:
@codex review

leoparente:
We decided to remove Corrade support and move forward with only statically linked handlers/inputs

PR #774, opened 2026-05-18
test(netprobe): cover process_netprobe_icmp via handcrafted IcmpLayers

leoparente:
Picks up where #772 left off. \`NetProbeMetricsManager::process_netprobe_icmp\` was the largest remaining gap in \`NetProbeStreamHandler.cpp\` after that PR's NetProbe unit tests — it was previously only reached through the network-dependent (now \`SKIP\`'d) parent tests.

pcap++ exposes \`pcpp::IcmpLayer::setEchoRequestData\` / \`setEchoReplyData\` for building a fully-formed ICMP layer in memory, no parsed pcap needed.

## Three new TEST_CASEs

| Case | Flow | Asserts |
|---|---|---|
| request/reply transaction lifecycle | REQUEST id=0x1234 seq=1 → REPLY id=0x1234 seq=1 within TTL | attempts=1, successes=1 |
| reply that times out records a failure | REQUEST → REPLY 6s later (exceeds default 5s TTL) | attempts=1, packets_timeout=1 |
| reply with no prior request is ignored | bare REPLY only (no matching REQUEST) | no target row created |

Together these cover the ECHO_REQUEST branch, both ECHO_REPLY result paths (Valid + TimedOut), and the implicit NotExist no-op — the entire body of \`process_netprobe_icmp\`.

## Out of scope

The original ask also called out \`NetflowData.h\` / \`SflowData.h\` (NetFlow v1+v7 PCAPs, sFlow IPv6 samples). Those need new wire-format fixtures (either hand-written byte sequences or a scapy-style generator script). Tracking as a separate task — not bundled here to keep this reviewable.

### DIFF
--- a/src/handlers/netprobe/test_net_probe.cpp
+++ b/src/handlers/netprobe/test_net_probe.cpp
@@ -1,5 +1,6 @@
 #include <catch2/catch_test_macros.hpp>
 #include <catch2/catch_test_visor.hpp>
+#include <catch2/otel_helpers.hpp>
 
 #include <sstream>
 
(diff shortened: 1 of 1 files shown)


leoparente:
@codex review

leoparente:
@codex review

samiura (approved):

Issue #109, opened 2021-07-06
RDRAND failure on Arista switch


weyrick:
Exception on startup. Both docker and appimage builds

weyrick:
```
[2021-07-06 17:40:28.832] [pktvisor] [info] pktvisor 3.2.1 starting up
[2021-07-06 17:40:28.832] [pktvisor] [info] Load input plugin: PcapInputModulePlugin dev.visor.module.input/1.0
[2021-07-06 17:40:28.832] [pktvisor] [info] Registering GET /api/v1/inputs/pcap/(\w+)
[2021-07-06 17:40:28.833] [pktvisor] [info] Load handler plugin: DnsHandler dev.visor.module.handler/1.0
[2021-07-06 17:40:28.833] [pktvisor] [info] Registering GET /api/v1/inputs/pcap/(\w+)/handlers/dns/(\w+)
[2021-07-06 17:40:28.833] [pktvisor] [info] Registering GET /api/v1/inputs/pcap/(\w+)/handlers/dns/(\w+)/bucket/(\d+)
[2021-07-06 17:40:28.833] [pktvisor] [info] Load handler plugin: NetHandler dev.visor.module.handler/1.0
[2021-07-06 17:40:28.833] [pktvisor] [info] Registering GET /api/v1/inputs/pcap/(\w+)/handlers/net/(\w+)
[2021-07-06 17:40:28.833] [pktvisor] [info] Registering GET /api/v1/inputs/pcap/(\w+)/handlers/net/(\w+)/bucket/(\d+)
[2021-07-06 17:40:28.833] [pktvisor] [info] Load handler plugin: PcapHandler dev.visor.module.handler/1.0
[2021-07-06 17:40:28.833] [pktvisor] [info] Registering GET /api/v1/inputs/pcap/(\w+)/handlers/pcap/(\w+)
[2021-07-06 17:40:28.834] [pktvisor] [info] Registering GET /api/v1/inputs/pcap/(\w+)/handlers/pcap/(\w+)/bucket/(\d+)
[2021-07-06 17:40:28.834] [pktvisor] [info] Initialize server control plane
[2021-07-06 17:40:28.834] [pktvisor] [info] Registering GET /api/v1/metrics/app
[2021-07-06 17:40:28.834] [pktvisor] [info] Registering GET /api/v1/metrics/rates
[2021-07-06 17:40:28.834] [pktvisor] [info] Registering GET /api/v1/metrics/bucket/(\d+)
[...]

PR #786, opened 2026-06-23
chore(metrics): replace vendored JSF PRNG with inline SplitMix64 (#700)

### LINKED ISSUE #700: [Proposal] Replace JSF with PCG-CPP
JSF was introduced in: https://github.com/orb-community/pktvisor/pull/110

https://github.com/imneme/pcg-cpp can be used through conan (https://conan.io/center/recipes/pcg-cpp?version=cci.20220409) which would reduce code being maintained on pktvisor side.

leoparente:
Closes #700.

## What
Replaces the vendored `3rd/rng/jsf.h` (178-line Bob Jenkins "Small Fast" PRNG) with a ~6-line inline **SplitMix64** in `AbstractMetricsManager.h`, and deletes the `3rd/rng/` directory.

JSF was used in exactly one spot — the deep-sampling gate:
```cpp
_deep_sampling_now.store((_rng() % 100U < _deep_sample_rate), ...);
```
i.e. "roll 0–99, deep-sample this event if the roll is under `deep_sample_rate`." The sampling **rate is unchanged**; only *which* events get deep-sampled differs, and nothing depends on the specific set (default `deep_sample_rate=100` means the RNG is never even called).

## Why SplitMix64 instead of the proposed pcg-cpp
The issue suggested the `pcg-cpp` conan package. I went a different way after weighing the options:
- **pcg-cpp adds a third-party dependency** to remove a dependency-free header — the wrong direction for "reduce maintained code" — and its idiomatic seeding `seed_seq_from<std::random_device>` is exactly the footgun that caused #109 (below).
- **`std::mt19937`** works but carries ~2.5 KB of state per handler instance for a 1-in-100 coin flip.
- **Inline SplitMix64** removes the whole `3rd/rng/` dir with **no new dependency**, 8-byte state, and BigCrush-grade quality. We "own" ~6 auditable lines instead of 178.

## Invariant preserved (#109)
#109 ("RDRAND failure on Arista switch") was a startup crash: the pre-JSF `randutils` RNG auto-seeded from `std::random_device` → `RDRAND`, which faults and throws on some hardware. PR #110 fixed it by switching to a **fixed-seeded** PRNG. SplitMix64 here is **seeded from a fixed constant and never from hardware entropy**, preserving that fix. A prominent comment documents the constraint so it isn't "improved" into entropy seeding later.

## Build wiring
- Delete `3rd/rng/` + its `add_subdirectory` + the `rng` link in `visor-core`.
- **Fix a latent copy-paste bug** in `3rd/timer/CMakeLists.txt`: it called `target_include_directories(rng ...)` instead of `(timer ...)`, so `timer.hpp`'s include dir was being attached to the `rng` target and only reached `visor-core` transitively through it. Removing `rng` would have broken `<timer.hpp>`; pointing it at `timer` makes the timer target self-contained.
- Update the `NOMINMAX` comment (it cited jsf.h's `static min()/max()`); the define stays for general Windows hygiene.

## Testing
`unit-tests-visor-core` (which includes the sampling/metrics tests) passes — 1153 assertions, 21 test cases.


leoparente:
@codex review

samiura (approved):

PR #791, opened 2026-06-26
refactor(metrics): single source of truth for summary quantile ranks/labels

leoparente:
## Problem

The SLA quantiles pktvisor reports (`0.50`, `0.90`, `0.95`, `0.99`) were restated in four places in `src/Metrics.h`, each positionally coupled to the order `get_quantiles()` produces:

- the `get_quantiles()` helper (the rank list)
- `to_json()` — `p50`/`p90`/`p95`/`p99` keys assigned to `quantiles[0..3]`
- `to_prometheus()` — four hand-written `LabelMap`s `"0.5"`/`"0.9"`/`"0.95"`/`"0.99"`
- `to_opentelemetry()` — a separate `fractions[4]{0.50,0.90,0.95,0.99}`

Nothing tied the emitted **label** to the **value** except array position, so changing or reordering the ranks at one site would silently mislabel the values at another. (This tidies up the `get_quantiles()` helper introduced when the bulk `kll_sketch::get_quantiles(fractions, n)` API was removed in the datasketches 5.x upgrade.)

## Change

Introduce one ordered source of truth and drive every path from it:

```cpp
struct QuantileSpec { double rank; const char *json_key; const char *prom_label; };
static constexpr std::array<QuantileSpec, 4> QUANTILES{{
    {0.50, "p50", "0.5"}, {0.90, "p90", "0.9"}, {0.95, "p95", "0.95"}, {0.99, "p99", "0.99"},
}};
```

- `get_quantiles()` loops the table.
- `to_json` / `to_prometheus` / `to_opentelemetry` loop the table for their key/label/rank, so values and labels can never desync.
- The table carries each representation explicitly (the rank `0.50` maps to the Prometheus label `"0.5"`, while `0.95` maps to `"0.95"`), so emitted strings are preserved exactly rather than re-derived by formatting.

Single file (`src/Metrics.h`). Also fixes a `quatile` typo in the helper param.

## Behavior

**No output change** — JSON keys, Prometheus `quantile` labels, and OpenTelemetry `set_quantile` ranks are byte-identical, in the same order, including the `_quantiles_sum` (SUM-merge) path.

## Testing

`unit-tests-visor-core` (sketches + metrics) and the handler suites that snapshot summary output — `unit-tests-handler-{net,net-v2,dns,dns-v2,flow,netprobe}` — all pass (7/7), confirming identical output.

### DIFF
--- a/src/Metrics.h
+++ b/src/Metrics.h
@@ -25,6 +25,7 @@
 #ifdef __GNUC__
 #pragma GCC diagnostic pop
 #endif
+#include <array>
 #include <chrono>
 #include <math.h>
 #include <regex>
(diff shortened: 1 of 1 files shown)


leoparente:
@codex review

samiura (approved):

